PR #13, opened 2020-02-12
[#170127598] Use the extended service model that includes metadata

alexgpeppe:
This PR aims to use the extended service model that includes also the services metadata.

### DIFF
--- a/utils/conversions.ts
+++ b/utils/conversions.ts
@@ -41,7 +41,7 @@ export function apiServiceToService(service: ApiService): Service {
 }
 
 // Returns an API Service Metadata from an internal Service model
-function getApiServiceMetadataFromRetrievedService(
+function toApiServiceMetadata(
   retrievedService: RetrievedService
 ): ApiServiceMetadata {
   return (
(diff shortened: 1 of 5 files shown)


alexgpeppe:
The changes depend on the `1.8.0` version of `io-functions-commons`, which has not been published yet. After this version is published, it's needed to update the dependency in the `package.json` file.

utils/conversions.ts
```diff
@@ -22,10 +23,44 @@ export function apiServiceToService(service: ApiService): Service {
...
+      scope: service.service_metadata.scope,
+      tosUrl: service.service_metadata.tos_url,
+      webUrl: service.service_metadata.web_url
+    },
     serviceName: service.service_name
   };
 }
 
+// Returns an API Service Metadata from an internal Service model
+function getApiServiceMetadataFromRetrievedService(
```
gunzip: `toApiServiceMetadata` (the "from" comes from the type of the first parameter)

utils/conversions.ts
```diff
@@ -22,10 +23,44 @@ export function apiServiceToService(service: ApiService): Service {
...
+  retrievedService: RetrievedService
+): ApiServiceMetadata {
+  return (
+    retrievedService.serviceMetadata && {
```
gunzip: this method returns the boolean "false" in case retrievedService.serviceMetadata is undefined, you should use a ternary operator here

digitalcitizenship:
## Affected stories

  * 🌟 [#170127598](https://www.pivotaltracker.com/story/show/170127598): Come backend voglio poter salvare le informazioni sui servizi in modo che possano essere utilizzate da IO

<p align="right">
  Generated by :no_entry_sign: <a href="http://github.com/danger/danger-js/">dangerJS</a>
</p>

PR #24, opened 2020-03-02
[#171536663] Several pipeline improvements

raicastino:
The following changes have been implemented to improve the current Azure DevOps pipeline for this repository:

- moved all the build packaging tasks from _Build_ stage (always executed) to _Deploy_x_ stages to **accelerate** the pipeline execution when deployment is skipped (i.e. not required)
- added **templates** usage to allow the reuse of several recurrent steps in the pipeline (e.g. install node modules and build, deploy a function, etc.): this improve readability, maintenance and reuse of content, logic, and parameters
- reviewed the conditions and renamed the pipeline **variables** to support the manual execution of specific jobs on-demand (e.g. allow administrators to execute the deployment of a specified branch or commit in production regardless ‘latest’ tag)
- modified the **deployment to slots** (optional alternative deployment mode enabled setting appropriate variables) to support deployment slots both in stage and production environment
- improved and added additional comments in the YML pipeline file to better **document** the pipeline

Screenshot 2020-03-02 at 12 11 37

### DIFF
--- a/.artifactignore
+++ b/.artifactignore
@@ -1,7 +1,0 @@
-*.js.map
-*.ts
-.vscode
-.git
-local.settings.json
-test
-tsconfig.json
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,34 +1,39 @@
 # Azure DevOps pipeline to build, check source codes, run tests, and deploy.
 #
-# The following variables needs to be defined before running the deployment jobs 
-# defined in this pipeline:
+# To enable the check of source code with Danger JS you need to configure a valid
+# GitHub token by setting the following variable: 
+# - DANGER_GITHUB_API_TOKEN
+#
+# To enable the deployment in any environment you need to configure the following 
+# variable otherwise all the deployment jobs will be always skipped:
+# - DO_DEPLOY = true
+# in case of a manual run, you also need to configure the following additional 
+# variables based on the environment to update:
+# - STAGING_ENABLE_MANUAL_DEPLOY = true
+# - PRODUCTION_ENABLE_MANUAL_DEPLOY = true
+#
+# The following variables are also used when running the deployment jobs:
+# - STAGING_DEPLOY_MODE: 'deploy_standard' || 'deploy_to_slots'
 # - STAGING_AZURE_SUBSCRIPTION
 # - STAGING_RESOURCE_GROUP_NAME
 # - STAGING_FUNCTION_APP_NAME
+# 
+# - PRODUCTION_DEPLOY_MODE: 'deploy_standard' || 'deploy_to_slots'
 # - PRODUCTION_AZURE_SUBSCRIPTION
 # - PRODUCTION_RESOURCE_GROUP_NAME
 # - PRODUCTION_FUNCTION_APP_NAME
 #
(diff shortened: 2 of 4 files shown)


.artifactignore
```diff
@@ -1,7 +0,0 @@
-*.js.map
```
gunzip: what's the reason to remove this file ?

azure-pipelines.yml
```diff
@@ -1,34 +1,40 @@
 # Azure DevOps pipeline to build, check source codes, run tests, and deploy.
 #
-# The following variables needs to be defined before running the deployment jobs 
-# defined in this pipeline:
+# To enable the check of source code with Danger JS you need to configure a valid
+# GitHub token by setting the following variable:
```
gunzip: do we know the the required permission for this token ? (repo:*) ?

azure-pipelines.yml
```diff
@@ -1,34 +1,40 @@
...
-# The following variables needs to be defined before running the deployment jobs 
-# defined in this pipeline:
+# To enable the check of source code with Danger JS you need to configure a valid
+# GitHub token by setting the following variable: 
+# - DANGER_GITHUB_API_TOKEN
+#
+# To enable the deployment in any environment you need to configure the following 
+# variable otherwise all the deployment jobs will be always skipped:
+# - DO_DEPLOY = true
+# in case of a manual run, you also need to configure the following additional
```
gunzip: how's a manual run started ?

.artifactignore
raicastino: @gunzip That file is (only) used in a Azure DevOps task for publishing the artefacts in Azure DevOps (I added it the previous pipeline)... In the new pipeline I removed this task and filtered the files to include/exclude in another way. So, it is no more used.

azure-pipelines.yml
raicastino: Based to the documentation for public repository (as all IO repos) it is only necessary to give the account token the smallest scope possible (this means just _public_repo_). 

In the current Azure DevOps pipeline, I configured the variable _DANGER_GITHUB_API_TOKEN_ with the token that was provided to me by @cloudify (I do not know the associated permissions).

azure-pipelines.yml
raicastino: @gunzip It is possible to use the Azure DevOps portal (the procedure will be documented), pressing the "Run pipeline" button after selecting the pipeline to run:

Screenshot 2020-03-02 at 14 05 34

Screenshot 2020-03-02 at 14 07 07

you have the option to select the DevOps stage(s) to run (e.g. only "Deploy_production") and override the preconfigured values of one or more variables when (manually) running the pipeline (e.g. _PRODUCTION_ENABLE_MANUAL_DEPLOY = true_). 

**Note.** I defined additional variables and conditions in the _azure-pipelines.yml_ file to be sure that a manual run of the pipeline will execute deployment tasks in staging/production environment only when explicitly setting those variables (to avoid unwanted executions by mistake).

gunzip: can you please put all these informations into some documents in the [io-handbook](https://github.com/pagopa/io-handbook) ?

PR #28, opened 2020-03-09
[#171549167] Add function to update the group list of a specific user

alexgpeppe:
This PR aims to add a function to edit the list of groups associated with a specific user from the API management.

### DIFF
--- a/UpdateUserGroups/handler.ts
+++ b/UpdateUserGroups/handler.ts
@@ -5,6 +5,7 @@ import * as express from "express";
 import { sequenceT } from "fp-ts/lib/Apply";
 import { array } from "fp-ts/lib/Array";
 import { either, toError } from "fp-ts/lib/Either";
+import { StrMap } from "fp-ts/lib/StrMap";
 import {
   fromEither,
   fromPredicate,
@@ -129,35 +130,35 @@ interface IGroupsClusterization {
 }
 
 function clusterizeGroups(
-  existingGroups: { [groupDisplayName: string]: string },
+  existingGroups: Record<string, string>,
   currentUserGroups: ReadonlyArray<string>,
   groupsInPayload: ReadonlyArray<string>
 ): IGroupsClusterization {
-  return Object.keys(existingGroups).reduce<IGroupsClusterization>(
-    (cluster, group) => {
+  return new StrMap(existingGroups).reduceWithKey(
+    { toBeAssociated: [], toBeRemoved: [] },
+    (displayName, cluster, name) => {
       if (
-        currentUserGroups.includes(group) &&
-        !groupsInPayload.includes(group)
+        currentUserGroups.includes(displayName) &&
+        !groupsInPayload.includes(displayName)
       ) {
         return {
           toBeAssociated: cluster.toBeAssociated,
-          toBeRemoved: cluster.toBeRemoved.concat([existingGroups[group]])
+          toBeRemoved: cluster.toBeRemoved.concat([name])
         };
       }
       if (
-        !currentUserGroups.includes(group) &&
-        groupsInPayload.includes(group)
+        !currentUserGroups.includes(displayName) &&
+        groupsInPayload.includes(displayName)
       ) {
         return {
           toBeAssociated: cluster.toBeAssociated.concat([
-            existingGroups[group]
+            existingGroups[displayName]
           ]),
           toBeRemoved: cluster.toBeRemoved
         };
       }
       return cluster;
-    },
-    { toBeAssociated: [], toBeRemoved: [] }
+    }
   );
 }
 
@@ -246,7 +247,7 @@ export function UpdateUserGroupHandler(
           .map(groupList => ({
             apimClient: taskResults.apimClient,
             currentUserGroups: taskResults.currentUserGroups,
-            existingGroups: groupList.reduce<{ [displayName: string]: string }>(
+            existingGroups: groupList.reduce<Record<string, string>>(
               (prev, curr) => ({ ...prev, [curr.displayName]: curr.name }),
               {}
             ),
--- a/UpdateUserGroups/handler.ts
+++ b/UpdateUserGroups/handler.ts
@@ -123,6 +123,44 @@ function getUserGroups(
   }, toError);
 }
 
+interface IGroupsClusterization {
+  toBeAssociated: ReadonlyArray<string>;
+  toBeRemoved: ReadonlyArray<string>;
+}
+
+function clusterizeGroups(
+  existingGroups: { [groupDisplayName: string]: string },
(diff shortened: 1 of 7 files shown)


UpdateUserGroups/handler.ts
```diff
@@ -0,0 +1,381 @@
...
+    while (nextLink) {
+      const nextGroupList = await apimClient.group.listByServiceNext(nextLink);
+      groupList.push(...nextGroupList);
+      nextLink = nextGroupList.nextLink;
+    }
+    return groupList;
+  }, toError);
+}
+
+function getUserGroups(
```
gunzip: isn't this the same code as in `GetUser` ? Can you put these methods into some shared module ?

UpdateUserGroups/handler.ts
```diff
@@ -0,0 +1,381 @@
...
+  toBeAssociated: ReadonlyArray<string>;
+  toBeRemoved: ReadonlyArray<string>;
+}
+
+function clusterizeGroups(
+  existingGroups: Record<string, string>,
+  currentUserGroups: ReadonlyArray<string>,
+  groupsInPayload: ReadonlyArray<string>
+): IGroupsClusterization {
+  return new StrMap(existingGroups).reduceWithKey(
```
gunzip: Maybe the `partion` method of StrMap could help here instead of reduce (or partitionMap)

UpdateUserGroups/handler.ts
```diff
@@ -0,0 +1,381 @@
...
+          azureApimConfig.apim,
+          taskResults.userName
+        )
+          .mapLeft(error =>
+            internalErrorHandler("Could not get the user groups.", error)
+          )
+          .map(currentUserGroups => ({
+            apimClient: taskResults.apimClient,
+            currentUserGroups: currentUserGroups.map(
+              groupContract => groupContract.displayName
```
gunzip: can you add some comments on why we use displayName instead of name

UpdateUserGroups/handler.ts
```diff
@@ -0,0 +1,381 @@
...
+    const groupList: GroupContract[] = [];
+    const groupListResponse = await apimClient.userGroup.list(
+      apimResourceGroup,
+      apim,
+      userName
+    );
+    groupList.push(...groupListResponse);
+    // tslint:disable-next-line:no-let
+    let nextLink = groupListResponse.nextLink;
+    while (nextLink) {
```
gunzip: not strictly needed for this pr but when you want to accumulate vaules another option is to use [unfoldr](https://github.com/gcanti/fp-ts/blob/1.12.0/src/Array.ts#L227) which is a do ... while that takes a function as an input that returns an Option (some -> put the value into the array and continue, none -> stop the loop)

UpdateUserGroups/handler.ts
alexgpeppe: I'm not sure about this... In the current implementation, only some of the existing groups are added to the clusterization, that is the groups with which the user must be associated and the groups from which the user must be deleted: if the user was not associated with a group and still must not be associated, then the group isn't included into the clusterization; the same for a group with which the user is already associated and with which must still be associated. If I use the partition `method`, instead, all the groups will be added to the clusterization, so an operation from the apim client will be performed for each of them even if it this is not necessary.

openapi/index.yaml
```diff
@@ -383,6 +412,25 @@ definitions:
     $ref: "https://raw.githubusercontent.com/teamdigitale/io-functions-commons/master/openapi/definitions.yaml#/OrganizationFiscalCode"
   ExtendedProfile:
     $ref: "https://raw.githubusercontent.com/teamdigitale/io-functions-commons/master/openapi/definitions.yaml#/ExtendedProfile"
+  UserGroupsPayload:
```
gunzip: can we add a description and/or an example with the expected format for this array ?

UpdateUserGroups/handler.ts
alexgpeppe: Thanks for the information! By the way, here we are accumulating values that come from a remote call using `await`; is there an equivalent method that allows using an asynchronous function? It seems I didn't see any in the docs...

BurnedMarshal:
# [Codecov](https://codecov.io/gh/pagopa/io-functions-admin/pull/28?src=pr&el=h1) Report
> Merging [#28](https://codecov.io/gh/pagopa/io-functions-admin/pull/28?src=pr&el=desc) into [master](https://codecov.io/gh/pagopa/io-functions-admin/commit/a1edb078321ee4a60d880f6c839272b1f35ab6ef&el=desc) will **increase** coverage by `0.79%`.
> The diff coverage is `92.72%`.

[Impacted file tree graph](https://codecov.io/gh/pagopa/io-functions-admin/pull/28?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master      #28      +/-   ##
==========================================
+ Coverage   87.07%   87.87%   +0.79%     
==========================================
  Files          19       20       +1     
  Lines         503      602      +99     
  Branches       30       36       +6     
==========================================
+ Hits          438      529      +91     
- Misses         64       72       +8     
  Partials        1        1              
```


| [Impacted Files](https://codecov.io/gh/pagopa/io-functions-admin/pull/28?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [GetUser/handler.ts](https://codecov.io/gh/pagopa/io-functions-admin/pull/28/diff?src=pr&el=tree#diff-R2V0VXNlci9oYW5kbGVyLnRz) | `91.17% <ø> (-1.23%)` | :arrow_down: |
| [UpdateUserGroups/handler.ts](https://codecov.io/gh/pagopa/io-functions-admin/pull/28/diff?src=pr&el=tree#diff-VXBkYXRlVXNlckdyb3Vwcy9oYW5kbGVyLnRz) | `91.83% <91.83%> (ø)` | |
| [utils/apim.ts](https://codecov.io/gh/pagopa/io-functions-admin/pull/28/diff?src=pr&el=tree#diff-dXRpbHMvYXBpbS50cw==) | `100.00% <100.00%> (ø)` | |

------

[Continue to review full report at Codecov](https://codecov.io/gh/pagopa/io-functions-admin/pull/28?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/pagopa/io-functions-admin/pull/28?src=pr&el=footer). Last update [a1edb07...113e0f0](https://codecov.io/gh/pagopa/io-functions-admin/pull/28?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

PR #39, opened 2020-03-30
[#171683040] administrations update function

alexgpeppe:


alexgpeppe:
@gunzip, there is currently a synchronization problem: the Activity that reads the file from indicePA seems not to get completed, the Orchestrator keeps waiting for its result.

PR #41, opened 2020-05-08
[#172724206] upgrade dependencies

balanza:
Upgrade of `io-functions-commons` from v4 to v9.

Consequently, custom telemetry implementation has been removed


package.json
```diff
@@ -9,7 +9,7 @@
     "prestart": "npm run build && func extensions install",
     "start:host": "func start",
     "start": "npm-run-all --parallel start:host watch",
-    "pretest": "yarn generate:definitions",
+    "_pretest": "yarn generate:definitions",
```
gunzip, line 12: why this change here ?

balanza: just a typo..

PR #42, opened 2020-05-11
[#172621882] Get all data for a specific user

balanza:
Implementation of part of the Activit handles needed to complete the task.

Things to note:
* some feature may fit better into `io-functions-commons`, like mocks. We might move them there after this activity
* `NotificationModel` has been extended to support a specific use case. This as well might be moved into commons.

### DIFF
--- a/ExtractUserDataActivity/handler.ts
+++ b/ExtractUserDataActivity/handler.ts
@@ -398,25 +398,28 @@ export const createExtractUserDataActivityHandler = (
         });
       });
 
-  // the actual handler
+  // the actual handler©
   return (context: Context, input: unknown) =>
-    fromEither(ActivityInput.decode(input))
-      .mapLeft<ActivityResultFailure>((reason: t.Errors) =>
-        ActivityResultInvalidInputFailure.encode({
-          kind: "INVALID_INPUT_FAILURE",
-          reason: readableReport(reason)
-        })
+    fromEither(
+      ActivityInput.decode(input).mapLeft<ActivityResultFailure>(
+        (reason: t.Errors) =>
+          ActivityResultInvalidInputFailure.encode({
+            kind: "INVALID_INPUT_FAILURE",
+            reason: readableReport(reason)
+          })
       )
+    )
       .chain(({ fiscalCode }) => queryAllUserData(fiscalCode))
-      .map(allUserData =>
-        ActivityResultSuccess.encode({
-          kind: "SUCCESS",
-          value: allUserData
-        })
+      .bimap(
+        failure => {
+          logFailure(context)(failure);
+          return failure;
+        },
+        allUserData =>
+          ActivityResultSuccess.encode({
+            kind: "SUCCESS",
+            value: allUserData
+          })
       )
-      .mapLeft(failure => {
-        logFailure(context)(failure);
-        return failure;
-      })
       .run();
 };
--- a/ExtractUserDataActivity/handler.ts
+++ b/ExtractUserDataActivity/handler.ts
@@ -5,8 +5,8 @@
 
 import * as t from "io-ts";
 
-import { sequenceS } from "fp-ts/lib/Apply";
-import { Either, fromOption, left } from "fp-ts/lib/Either";
+import { sequenceS, sequenceT } from "fp-ts/lib/Apply";
+import { Either, fromOption, left, right } from "fp-ts/lib/Either";
 import {
   fromEither,
   TaskEither,
@@ -16,12 +16,22 @@ import {
 
 import { Context } from "@azure/functions";
 
-import { QueryError } from "documentdb";
+import { BlobService } from "azure-storage";
 import {
-  Message,
-  MessageModel
+  QueryError,
+  RetrievedDocument as RetrievedDocumentT
+} from "documentdb";
+import { MessageContent } from "io-functions-commons/dist/generated/definitions/MessageContent";
+import {
+  MessageModel,
+  MessageWithContent,
+  RetrievedMessageWithContent,
+  RetrievedMessageWithoutContent
 } from "io-functions-commons/dist/src/models/message";
-import { Notification } from "io-functions-commons/dist/src/models/notification";
+import {
+  Notification,
+  RetrievedNotification
+} from "io-functions-commons/dist/src/models/notification";
 import {
   Profile,
   ProfileModel
@@ -37,7 +47,7 @@ import { NotificationModel } from "./notification";
 
 // the shape of the dataset to be extracted
 const AllUserData = t.interface({
-  messages: t.readonlyArray(Message, "MessageList"),
+  messages: t.readonlyArray(MessageWithContent, "MessageList"),
   notifications: t.readonlyArray(Notification, "NotificationList"),
   profile: Profile,
   senderServices: t.readonlyArray(SenderService, "SenderServiceList")
@@ -110,16 +120,16 @@ const logPrefix = `ExtractUserDataActivity`;
  * @returns either the query result or a query failure
  */
 const fromQueryEither = <R>(
-  lazyPromise: () => Promise<Either<QueryError, R>>,
+  lazyPromise: () => Promise<Either<QueryError | Error, R>>,
   queryName: string = ""
-) =>
+): TaskEither<ActivityResultQueryFailure, R> =>
   tryCatch(lazyPromise, (err: Error) =>
     ActivityResultQueryFailure.encode({
       kind: "QUERY_FAILURE",
       query: queryName,
       reason: err.message
     })
-  ).chain((queryErrorOrRecord: Either<QueryError, R>) =>
+  ).chain((queryErrorOrRecord: Either<QueryError | Error, R>) =>
     fromEither(
       queryErrorOrRecord.mapLeft(queryError =>
         ActivityResultQueryFailure.encode({
@@ -131,6 +141,10 @@ const fromQueryEither = <R>(
     )
   );
 
+/**
+ * To be used for exhaustive checks
+ * @param _
+ */
 function assertNever(_: never): void {
   throw new Error("should not have executed this");
 }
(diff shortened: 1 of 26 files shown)


SetUserDataProcessingStatusActivity/handler.ts
```diff
@@ -0,0 +1,170 @@
...
+          ...currentRecord,
+          status: nextStatus
+        }),
+      (err: Error) => {
+        return ActivityResultQueryFailure.encode({
+          kind: "QUERY_FAILURE",
+          query: "userDataProcessingModel.createOrUpdateByNewOne",
+          reason: err.message
+        });
+      }
```
AleDore, line 137: Can we use `fromQueryEither` in this case?

balanza: I agree, but still `fromQueryEither` isn't as generic as to be placed in a dedicated module. I'll just copy it over though, because I think it's more eloquent

ExtractUserDataActivity/notification.ts
```diff
@@ -0,0 +1,51 @@
...
+      this.dbClient,
+      this.collectionUri,
+      {
+        parameters: [
+          {
+            name: "@fiscalCode",
+            value: fiscalCode
+          }
+        ],
+        query: `SELECT * FROM m WHERE m.fiscal_code = @fiscalCode`
```
gunzip, line 46: since the partitionKey of the notifications collection is `messageId` this results in a cross partition query which should be avoided (see https://docs.microsoft.com/it-it/azure/cosmos-db/how-to-query-container#cross-partition-query). 

you should include a contidion on `messageId` so: get all users messages then, for each retrieved message, loop to get the message-status and the relative nofitications; then, for each retrieved notification, loop to get then the notification-status.

SetUserDataProcessingStatusActivity/handler.ts
```diff
@@ -0,0 +1,184 @@
...
+ *
+ * The validation process use a `id and validator` strategy.
+ *
+ * The `id` is generated using an ulid generator and is used when searching
+ * a specific ValidationToken entity in the table storage.
+ *
+ * For the `validator` we use a random-bytes generator. This `validator` value is
+ * hashed using the `sha256` strategy and then stored in the entity as `validatorHash`
+ *
+ * Each token has also a `InvalidAfter` field to set the token lifetime.
```
gunzip, line 12: probably we should remove this comments? :)

ExtractUserDataActivity/handler.ts
```diff
@@ -0,0 +1,526 @@
...
+  }
+};
+
+/**
+ * Skims a db document from db-related fields
+ * @param doc the document as retrieved from db
+ *
+ * @returns the same document without db metadata
+ */
+const fromRetrievedDbDocument = <T>(doc: T & RetrievedDocumentT): T => {
```
gunzip, line 194: every model has a class method `toBaseType` method which does exactly the same and does not depend on cosmosdb internal, consider using that one and remove this one
ie. https://github.com/pagopa/io-functions-commons/blob/6b60b5cc5fb840ff397f2ebc3261a3c7408b2168/src/models/profile.ts#L129

ExtractUserDataActivity/handler.ts
```diff
@@ -0,0 +1,526 @@
...
+ * Safely compose a message with a messa with its content
+ * @param messageFromDb a message as it's stored in db
+ * @param content a content for the message
+ *
+ * @returns a new message with content included
+ */
+const appendContentToMessage = (
+  messageFromDb: RetrievedMessageWithoutContent,
+  content: MessageContent
+): MessageWithContent => ({
```
gunzip, line 219: why do we need to merge the message with its content ?
it's better to just save them in a separate blob

ExtractUserDataActivity/handler.ts
```diff
@@ -0,0 +1,526 @@
...
+            ...q,
+            [n.id, NotificationChannelEnum.EMAIL],
+            [n.id, NotificationChannelEnum.WEBHOOK]
+          ],
+          []
+        )
+        .map(([notificationId, channel]) =>
+          fromQueryEither(
+            () =>
+              notificationStatusModel.findOneNotificationStatusByNotificationChannel(
```
gunzip, line 414: why do we need to differentiate by channel?

ExtractUserDataActivity/handler.ts
```diff
@@ -0,0 +1,526 @@
...
+   * @param messages
+   */
+  const findNotificationsForAllMessages = (
+    messages: readonly RetrievedMessageWithoutContent[]
+  ): TaskEither<
+    ActivityResultQueryFailure,
+    ReadonlyArray<RetrievedNotification>
+  > => {
+    if (messages.length) {
+      // this spread is needed as typescript wouldn't recognize messages[0] to be defined otherwise
```
gunzip, line 358: it's not clear, why do we need this pattern here ? consider using a NonEmptyArray instead

NonEmptyArray.decode(messages)... or if (NonEmptyArray.is(message)) ...

gunzip: "fp-ts/lib/NonEmptyArray"

ExtractUserDataActivity/handler.ts
balanza: it seems to to be used by anyone. It's passed to the base class which exposes it as protected, then never actually used.
https://github.com/pagopa/io-functions-commons/blob/master/src/utils/documentdb_model.ts#L31

ExtractUserDataActivity/handler.ts
balanza: good, even easier

ExtractUserDataActivity/handler.ts
gunzip: If we cannot use that, I suggest to remove this method anyway and just use 
`t.exact(Message).encode(o);` inline when needed

ExtractUserDataActivity/handler.ts
gunzip: the structure of the zip should mimic the internal representation (that won't happen, but suppose you have to restore all data fom the content of the zip :)

ExtractUserDataActivity/handler.ts
balanza: no, but I wanted to use the actual method which queries by statusId, and calculates the statusId by its channel

ExtractUserDataActivity/handler.ts
balanza: This includes also document-related data? I mean: should we export also `RetrievedDocument`-related fields?

ExtractUserDataActivity/handler.ts
balanza: didn't know that, thanks

ExtractUserDataActivity/handler.ts
gunzip: no since they're not needed to rebuild the whole thing

ExtractUserDataActivity/handler.ts
gunzip: no since they're not needed to rebuild the whole thing

ExtractUserDataActivity/handler.ts
balanza: are you sure it works like that? it doesn't seem to be a decoder though.

Anyway, I think it doesn't solved the issue: to pass an arbitrary list of taskEither to a `sequenceT` to resolve them in parallel: https://github.com/pagopa/io-functions-admin/pull/42/files/98dd1f4ddb37bed424eb078bd0549cd0a9c064b9#diff-6c3441a7b004efdd3512746300508009R339

ExtractUserDataActivity/handler.ts
```diff
@@ -0,0 +1,526 @@
...
+              ),
+            "findNotificationsForRecipient"
+          )
+        )
+      ).foldTaskEither(
+        e => fromEither(left(e)),
+        arrayOfArray =>
+          fromEither(
+            right(
+              arrayOfArray.reduce(
```
gunzip, line 383: you can use array.flatten from fp-ts

ExtractUserDataActivity/handler.ts
gunzip: I mean: you can just use `NonEmptyArray.fromArray(messages).map(...)` instead of check messages length and separate head and tail (it's not clear why do you want to do that)

ExtractUserDataActivity/handler.ts
balanza: I think it doesn't support readonly arrays (introduced in v2.5)

ExtractUserDataActivity/handler.ts
```diff
@@ -0,0 +1,445 @@
...
+): void => {
+  switch (failure.kind) {
+    case "INVALID_INPUT_FAILURE":
+      context.log.error(
+        `${logPrefix}|Error decoding input|ERROR=${failure.reason}`
+      );
+      break;
+    case "QUERY_FAILURE":
+      context.log.error(
+        `${logPrefix}|Error ${failure.query} query error |ERROR=${failure.reason}`
```
gunzip, line 182:
```suggestion
        `${logPrefix}|Error ${failure.query} query error|ERROR=${failure.reason}`
```

ExtractUserDataActivity/handler.ts
```diff
@@ -0,0 +1,445 @@
...
+      );
+      break;
+    case "USER_NOT_FOUND_FAILURE":
+      context.log.error(`${logPrefix}|Error user not found |ERROR=`);
```
gunzip, line 186:
```suggestion
      context.log.error(`${logPrefix}|Error user not found|ERROR=`);
```

ExtractUserDataActivity/handler.ts
```diff
@@ -0,0 +1,445 @@
...
+                  messageId
+                })
+              )
+            )
+        )
+      )
+    );
+
+  /**
+   * Given a list of messages, it queires for relative notifications
```
gunzip, line 279:
```suggestion
   * Given a list of messages, get the relative notifications
```

ExtractUserDataActivity/handler.ts
```diff
@@ -0,0 +1,445 @@
...
+            "findNotificationsForRecipient"
+          )
+        )
+      )
+      .foldTaskEither(
+        e => fromEither(left(e)),
+        arrayOfArray =>
+          // tslint:disable-next-line: readonly-array
+          fromEither(right(flatten(arrayOfArray as RetrievedNotification[][])))
+      );
```
gunzip, line 305: since you are converting back to taskEither, maybe I'm wrong but it looks like you can just use `map(flatten)` here

ExtractUserDataActivity/handler.ts
```diff
@@ -0,0 +1,445 @@
...
+    ActivityResultQueryFailure,
+    ReadonlyArray<MessageContentWithId>
+  > =>
+    array.sequence(taskEither)(
+      messages.map(({ id: messageId }) =>
+        fromQueryEither(
+          () => messageModel.getContentFromBlob(blobService, messageId),
+          "messageModel.getContentFromBlob"
+        ).foldTaskEither<ActivityResultQueryFailure, MessageContentWithId>(
+          failure => fromEither(left(failure)),
```
gunzip, line 258: when the left value is unchanged, probably you can `chain` or `map` instead of folding

ExtractUserDataActivity/handler.ts
```diff
@@ -0,0 +1,445 @@
...
+          .reduce(
+            (queries, { id: notificationId }) => [
+              ...queries,
+              ...Object.values(NotificationChannelEnum).map(channel => {
+                switch (channel) {
+                  case NotificationChannelEnum.EMAIL:
+                  case NotificationChannelEnum.WEBHOOK:
+                    return [notificationId, channel];
+                  default:
+                    assertNever(channel);
```
gunzip, line 326: cannot `...Object.values(NotificationChannelEnum).map(channel => [notificationId, channel])` 
just work without the switch here?

ExtractUserDataActivity/handler.ts
```diff
@@ -0,0 +1,445 @@
...
+              "findOneNotificationStatusByNotificationChannel"
+            )
+          )
+      )
+      // filter empty results (it might not exist a content for a pair notification/channel)
+      .foldTaskEither<
+        ActivityResultQueryFailure,
+        ReadonlyArray<NotificationStatus>
+      >(
+        e => fromEither(left(e)),
```
gunzip, line 348: map / chain ?

ExtractUserDataActivity/handler.ts
```diff
@@ -0,0 +1,445 @@
...
+  return (context: Context, input: unknown) =>
+    fromEither(ActivityInput.decode(input))
+      .mapLeft<ActivityResultFailure>((reason: t.Errors) =>
+        ActivityResultInvalidInputFailure.encode({
+          kind: "INVALID_INPUT_FAILURE",
+          reason: readableReport(reason)
+        })
+      )
+      .chain(({ fiscalCode }) => queryAllUserData(fiscalCode))
+      .map(allUserData =>
```
gunzip, line 434: when you have `map` + `mapLeft` you can use `bimap` to handle both

ExtractUserDataActivity/index.ts
```diff
@@ -0,0 +1,89 @@
...
+const senderServiceModel = new SenderServiceModel(
+  documentClient,
+  documentDbUtils.getCollectionUri(
+    documentDbDatabaseUrl,
+    SENDER_SERVICE_COLLECTION_NAME
+  )
+);
+
+const blobService = createBlobService(
+  getRequiredStringEnv("StorageConnection")
```
gunzip, line 77:
```suggestion
  getRequiredStringEnv("UserDataArchiveStorageConnection")
```

StorageConnection is reserved for the function runtime

ExtractUserDataActivity/notification.ts
```diff
@@ -0,0 +1,68 @@
...
+   */
+  constructor(
+    dbClient: DocumentDb.DocumentClient,
+    collectionUrl: DocumentDbUtils.IDocumentDbCollectionUri
+  ) {
+    super(dbClient, collectionUrl);
+  }
+
+  /**
+   * Returns the messages for the provided fiscal code
```
gunzip, line 46:
```suggestion
   * Returns the notifications for the provided message id
```

ExtractUserDataActivity/handler.ts
balanza: Yes, but this way I could implement an exahustive check that will force us to update this function if in the future we add/remove options in `NotificationChannelEnum `

gunzip: why do we want to be forced to update this method in case?

ExtractUserDataActivity/handler.ts
balanza: you're right, `foldTaskEither` come from a version of the code which has been refactored now. No more reason for that, thanks

ExtractUserDataActivity/handler.ts
balanza: In this case it's not trivial, because the _right_ path may become _left_ if Option is none.

ExtractUserDataActivity/handler.ts
balanza: Maybe we'll add support for another notification channel

gunzip: and why `.map(channel => [notificationId, channel])` won't suffice in that case?

PR #43, opened 2020-05-14
[#172621882] generate zip file with user data

balanza:
The following implements the generation of a zip bundle with all user data, and save it to a storage.

The functionality is exposed as a runnable script for local execution

### Run locally
```sh
yarn build && node dist/ExtractUserDataActivity/cli.js <FISCAL CODE>
```

### Notable points
* bundles are compressed as zip and encrypted with `zip20` algorithm
* the activity returns either an error or a object containing the name of the saved blob and the password to open the zip

### Yet to be solved
* `AllUserData.encode` fails on actual db data. This is important as the encoder would purge any sensitive data. While this is unsolved, [a patch is applied](https://github.com/pagopa/io-functions-admin/compare/172621882-accesso-ai-dati?expand=1#diff-6c3441a7b004efdd3512746300508009R536). 
* `utils/zip` module should be unit tested

### DIFF
--- a/ExtractUserDataActivity/cli.ts
+++ b/ExtractUserDataActivity/cli.ts
@@ -1,5 +1,5 @@
 /**
- * Esposes ExtractUserDataActivity as a cli command for local usage
+ * Exposes ExtractUserDataActivity as a cli command for local usage
  */
 
 // tslint:disable: no-console
--- a/ExtractUserDataActivity/cli.ts
+++ b/ExtractUserDataActivity/cli.ts
@@ -1,0 +1,22 @@
+/**
+ * Esposes ExtractUserDataActivity as a cli command for local usage
+ */
+
+// tslint:disable: no-console
+import { Context } from "@azure/functions";
+import { readableReport } from "italia-ts-commons/lib/reporters";
+import { FiscalCode } from "italia-ts-commons/lib/strings";
+import activity from "./index";
+
+const fiscalCode = FiscalCode.decode(process.argv[2]).getOrElseL(reason => {
+  throw new Error(`Invalid input: ${readableReport(reason)}`);
+});
+
+const context = ({
+  log: console
+  // tslint:disable-next-line: no-any
+} as any) as Context;
+
+activity(context, { fiscalCode })
+  .then(result => console.log("OK", result))
+  .catch(ex => console.error("KO", ex));
--- a/ExtractUserDataActivity/handler.ts
+++ b/ExtractUserDataActivity/handler.ts
@@ -398,25 +398,28 @@ export const createExtractUserDataActivityHandler = (
         });
       });
 
-  // the actual handler
+  // the actual handler©
   return (context: Context, input: unknown) =>
-    fromEither(ActivityInput.decode(input))
-      .mapLeft<ActivityResultFailure>((reason: t.Errors) =>
-        ActivityResultInvalidInputFailure.encode({
-          kind: "INVALID_INPUT_FAILURE",
-          reason: readableReport(reason)
-        })
+    fromEither(
+      ActivityInput.decode(input).mapLeft<ActivityResultFailure>(
+        (reason: t.Errors) =>
+          ActivityResultInvalidInputFailure.encode({
+            kind: "INVALID_INPUT_FAILURE",
+            reason: readableReport(reason)
+          })
       )
+    )
       .chain(({ fiscalCode }) => queryAllUserData(fiscalCode))
-      .map(allUserData =>
-        ActivityResultSuccess.encode({
-          kind: "SUCCESS",
-          value: allUserData
-        })
+      .bimap(
+        failure => {
+          logFailure(context)(failure);
+          return failure;
+        },
+        allUserData =>
+          ActivityResultSuccess.encode({
+            kind: "SUCCESS",
+            value: allUserData
+          })
       )
-      .mapLeft(failure => {
-        logFailure(context)(failure);
-        return failure;
-      })
       .run();
 };
--- a/ExtractUserDataActivity/handler.ts
+++ b/ExtractUserDataActivity/handler.ts
@@ -5,8 +5,8 @@
 
 import * as t from "io-ts";
 
-import { sequenceS } from "fp-ts/lib/Apply";
-import { Either, fromOption, left } from "fp-ts/lib/Either";
+import { sequenceS, sequenceT } from "fp-ts/lib/Apply";
+import { Either, fromOption, left, right } from "fp-ts/lib/Either";
 import {
   fromEither,
   TaskEither,
@@ -16,12 +16,22 @@ import {
 
 import { Context } from "@azure/functions";
 
-import { QueryError } from "documentdb";
+import { BlobService } from "azure-storage";
 import {
-  Message,
-  MessageModel
+  QueryError,
+  RetrievedDocument as RetrievedDocumentT
+} from "documentdb";
+import { MessageContent } from "io-functions-commons/dist/generated/definitions/MessageContent";
+import {
+  MessageModel,
+  MessageWithContent,
+  RetrievedMessageWithContent,
+  RetrievedMessageWithoutContent
 } from "io-functions-commons/dist/src/models/message";
-import { Notification } from "io-functions-commons/dist/src/models/notification";
+import {
+  Notification,
+  RetrievedNotification
+} from "io-functions-commons/dist/src/models/notification";
 import {
   Profile,
   ProfileModel
@@ -37,7 +47,7 @@ import { NotificationModel } from "./notification";
 
 // the shape of the dataset to be extracted
 const AllUserData = t.interface({
-  messages: t.readonlyArray(Message, "MessageList"),
+  messages: t.readonlyArray(MessageWithContent, "MessageList"),
   notifications: t.readonlyArray(Notification, "NotificationList"),
   profile: Profile,
   senderServices: t.readonlyArray(SenderService, "SenderServiceList")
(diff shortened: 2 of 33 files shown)


ExtractUserDataActivity/cli.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Esposes ExtractUserDataActivity as a cli command for local usage
```
gunzip, line 2:
```suggestion
 * Exposes ExtractUserDataActivity as a cli command for local usage
```

ExtractUserDataActivity/handler.ts
```diff
@@ -432,6 +464,71 @@ export const createExtractUserDataActivityHandler = (
...
+          {
+            [fileName]: data
+          },
+          password
+        );
+
+        const writableBlobStream = blobService.createWriteStreamToBlockBlob(
+          userDataContainerName,
+          blobName,
+          (err, _) => {
```
gunzip, line 500: it's not clear why do we need the callback here ? cannot we just call 

` const writableBlobStream = blobService.createWriteStreamToBlockBlob(userDataContainerName, blobName)`

and write to the archiver stream using `append` ?

utils/random.ts
```diff
@@ -0,0 +1,51 @@
+/**
```
gunzip, line 1: consider using a package ie https://github.com/klughammer/node-randomstring instead of writing our custom implementation

utils/zip.ts
```diff
@@ -0,0 +1,36 @@
...
+import { NonEmptyString } from "italia-ts-commons/lib/strings";
+import { Readable } from "stream";
+archiver.registerFormat("zip-encrypted", achiverEncryptedFormat);
+
+export enum EncryptionMethodEnum {
+  ZIP20 = "zip20",
+  AES256 = "aes256"
+}
+
+export const ENCRYPTION_METHOD = "zip20";
```
gunzip, line 12:
```suggestion
export const DEFAULT_ENCRYPTION_METHOD = EncryptionMethodEnum.ZIP20;
```

gunzip: is this used anywhere?

utils/zip.ts
```diff
@@ -0,0 +1,36 @@
...
+  // tslint:disable-next-line: no-any
+  data: Record<string, any>,
+  password?: NonEmptyString,
+  encryptionMethod: EncryptionMethodEnum = EncryptionMethodEnum.ZIP20
+): Readable => {
+  const zipArchive = password
+    ? archiver("zip-encrypted", {
+        encryptionMethod,
+        password,
+        zlib: { level: 8 }
```
gunzip, line 24: I'd extract this value into DEFAULT_ZLIB_LEVEL

ExtractUserDataActivity/handler.ts
```diff
@@ -432,6 +464,71 @@ export const createExtractUserDataActivityHandler = (
...
+                null,
+                ArchiveInfo.encode({
+                  blobName,
+                  password
+                })
+              );
+            }
+          }
+        );
+        readableZipStream.pipe(writableBlobStream);
```
gunzip, line 519: pipe should be called before any operation on the stream (append / finalize)

ExtractUserDataActivity/handler.ts
```diff
@@ -189,8 +214,15 @@ export const createExtractUserDataActivityHandler = (
...
   senderServiceModel: SenderServiceModel,
-  blobService: BlobService
-  // tslint:disable-next-line: no-big-function
+  blobService: BlobService,
+  userDataContainerName: NonEmptyString,
+  createCompressedStream: (
+    // tslint:disable-next-line: no-any
+    data: Record<string, any>,
+    password: NonEmptyString
+  ) => stream.Readable
```
gunzip, line 223: it would have been useful to pass this as a parameter in case you'd used some generic typings 
(ie. () => stream.Readable) but since this is the exact type of `createCompressedStream` then you can just import the method and use it directly without passing it as a parameter

balanza: I did it mainly for testing, so it's easier to mock such function

gunzip: we'll find a way to mock it anyway ie `jest.mock("../zip")`

utils/random.ts
balanza: I will if we'll need more complex stuff, so far it's easy job and I'd rather not add another dependency

ExtractUserDataActivity/handler.ts
balanza: The callback refers to the saving on the remote storage ([docs](https://azure.github.io/azure-storage-node/BlobService.html#createWriteStreamToBlockBlob)).

We use `append` to add content as you'd expect, but the final result of the operation is given in the callback

utils/random.ts
gunzip: I'll do that instead, your code is unsafely using Math.random, we need something better:
https://github.com/klughammer/node-randomstring/blob/master/lib/randomstring.js#L9

more code = more bugs, please use randomstring

gunzip: bonus: is already included: https://github.com/pagopa/io-functions-admin/blob/master/package.json#L61

ExtractUserDataActivity/handler.ts
gunzip: that's right but the callback is optional (it will write on the remote storage even if you don't provide any). why do we need it ?

utils/random.ts
balanza: random string isn't working for us as it doesn't ensure the string will contain some for character for any character group. I'd need to fin another lib to include

ExtractUserDataActivity/handler.ts
balanza: because otherwise there would not be not a way to know if and when the blob has been saved to the storage. 
The flow is: _zipStream_ --> _blobStream_ --> _blobSaving_. We can listen for the `end` and `finish` event, but still the saving step can fail imho

utils/random.ts
gunzip: where does this requirement come from ? you can concatenate more random strings if you really need this (do we?)

balanza: "Passwords are needed to be strong". What does _strong_ mean? I don't know. My interpretation was:
* 18 characters
* at least one uppercase letter
* at least one lowercase letter
* at least one number
* at least one symbol
* no pattern
Shold we check with security guys?

ExtractUserDataActivity/handler.ts
balanza: It is _implicity_ true already as `append` and `finalize` from the `archiver` module operate on another eventloop iteration. I agree we should write explicit code 
I see three options:
1. we use the `archiver-append` function outside of the `zip` module, so we can call it after the zip stream has been piped. This leads to a leak of the archiver api to the activity handler, which may not know about it
2. we keep the `archiver-append` call inside the `zip` module, as it is now, but we enforce a `process.nextTick`
3. (don't know if it works) we keep the `archiver-append` call inside the `zip` module, but we wait for the `pipe` event of the readable stream (so we're sure we won't load files before it has been piped to somewhere)

Keep in mind that the main reason of having a `zip` module is that module initialization must happen _once per application_, otherwise it would throw.

utils/__mocks__/zip.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Mock implementation of the zip module
+ */
+
+export const createCompressedStream = jest.fn(() => ({
```
gunzip, line 5: generally we don't need this kind of mocks, simply call `jest.mock(<module path>)` will transform every exported functions form that module into a mock (jest.fn)

utils/random.ts
gunzip: according to pietro we don't need a constraint on the number of symbols for each set, so a 18 character string with one character set that has upper/lower case alphanumeric characters and special symbols is ok

utils/random.ts
```diff
@@ -0,0 +1,63 @@
...
+const STRONG_PASSWORD_PATTERN =
+  "(?=.{10,})(?=.*[!£$%&()?+@=€].*)(?=.*[a-z].*)(?=.*[A-Z].*)(?=.*[0-9].*)";
+
+export const StrongPassword = t.intersection([
+  PatternString(STRONG_PASSWORD_PATTERN),
+  NonEmptyString
+]);
+export type StrongPassword = t.TypeOf<typeof StrongPassword>;
+
+const shuffleString = (str: string): string => {
```
gunzip, line 25: see the comment above, remove this module and just generate a 18 character string from the needed charset

utils/zip.ts
```diff
@@ -0,0 +1,37 @@
...
+export const DEFAULT_ENCRYPTION_METHOD = EncryptionMethodEnum.ZIP20;
+export const DEFAULT_ZLIB_LEVEL = 8;
+
+export const createCompressedStream = (
+  // tslint:disable-next-line: no-any
+  data: Record<string, any>,
+  password?: NonEmptyString,
+  encryptionMethod: EncryptionMethodEnum = EncryptionMethodEnum.ZIP20
+): Readable => {
+  const zipArchive = password
```
gunzip, line 21: in which case we want a passwordless zip ? this is not needed.

ExtractUserDataActivity/handler.ts
gunzip: I'd simply put all the implementation here and remove the zip module. importing modules must not have side effects, even more if this may cause an exception. if you need to initialize something, implement some init() method and make it safe.

utils/zip.ts
balanza: > is this used anywhere?

Not actually

PR #45, opened 2020-05-15
[#172621882] small refactor: generate zip with user data

gunzip:
- removed useless code in zip module
- removed useless code in random module
- removed a wrong test


ExtractUserDataActivity/__tests__/handler.test.ts
```diff
@@ -125,36 +116,6 @@ describe("createExtractUserDataActivityHandler", () => {
...
-      aUserDataContainerName
-    );
-    const input: ActivityInput = {
-      fiscalCode: aFiscalCode
-    };
-
-    await handler(contextMock, input);
-
-    // @ts-ignore as
-    const mockCall = createCompressedStream.mock.calls[0];
```
gunzip, line 152: this is wrong and we must rewrite it. if you mock the blobService and call cb() immediately the code following `createWriteStreamToBlockBlob` won't never be executed

utils/zip.ts
```diff
@@ -1,37 +1,24 @@
...
-        zlib: { level: DEFAULT_ZLIB_LEVEL }
-      })
-    : archiver("zip", {
-        zlib: { level: DEFAULT_ZLIB_LEVEL }
-      });
-
-  Object.entries(data).forEach(([fileName, content]) => {
-    zipArchive.append(JSON.stringify(content), { name: fileName });
-  });
-  zipArchive.finalize();
```
gunzip, line 34: this is a promise and must be awaited somewhere

PR #47, opened 2020-05-16
[#172621882] small refactor: generate zip with user data

gunzip:
- handle archive (zip) promise
- fix unit tests


balanza:
@gunzip I agree on everything but the dismissal of the test about sensitive data (notification webhook url). I prepared an addendum on this branch in #48, I think we should merge that before merging this.

PR #48, opened 2020-05-16
[#172621882] refactor addendum: test no sensitive data is leaked into zip bundle

balanza:
This is an addendum to #47 that restores the test against bundled data, to prevent notification webhook urls to be included in the bundled zip


gunzip:
thank you, I've added the test here: https://github.com/pagopa/io-functions-admin/pull/47/files#diff-c986ac3863fd52a465579aadfb0884a0R175 !

gunzip:
FYI it was the last refactor atm, I'm swithing to work in send in-app message activity

PR #50, opened 2020-05-17
[#172621882] send a message to the user that requested her own data

gunzip:
draft as test must be refactored to handle promises (return value from activities) instead of taskeithers


ExtractUserDataActivity/handler.ts
```diff
@@ -583,5 +580,6 @@ export function createExtractUserDataActivityHandler(
             value: archiveInfo
           })
       )
-      .run();
+      .run()
+      .then(e => e.value);
```
gunzip, line 584: you cannot return an Either from an activity since the return value must be serializable

PR #52, opened 2020-05-18
[#172621882] use different blob service for message and user data

gunzip:


UserDataDownloadOrchestrator/cli.ts
```diff
@@ -5,7 +5,9 @@
 // tslint:disable: no-console no-any
 
 import * as dotenv from "dotenv";
-dotenv.config();
+dotenv.config({
+  path: ".env.test"
```
balanza, line 9: Why do we need this?

PR #53, opened 2020-05-18
[#172885142] delete user data cli command

balanza:
Includes:
* cli command
* redis session purge
* block/unblock user during the execution

### How to use
```sh
yarn build && REDIS_URL=redis://url-to-redis node dist/UserDataDeleteOrchestrator/cli.js <FISCAL_CODE>
```

### To implement yet
* ~change user data processing status~
* bundle user data and save it to a dedicated storage
* delete all user data


UserDataDeleteOrchestrator/session-utils/redisSessionStorage.ts
```diff
@@ -0,0 +1,263 @@
...
+   * {@inheritDoc}
+   */
+  private async del(
+    sessionToken: SessionToken,
+    walletToken: WalletToken
+  ): Promise<Either<Error, boolean>> {
+    const deleteSessionTokens = new Promise<Either<Error, true>>(resolve => {
+      // Remove the specified key. A key is ignored if it does not exist.
+      // @see https://redis.io/commands/del
+      this.redisClient.del(
```
gunzip, line 175: can you put some logging for every ops ? ie:

```
getting user sessions... <session1> ... <session2> ...
deleting SESSIONIFO-< token> ì ...
deleting USERMETA-<CF>
...
```

UserDataDeleteOrchestrator/cli.ts
```diff
@@ -0,0 +1,143 @@
...
+): TaskEither<Error, true> => notImplementedTask("deleteUserData");
+
+// change status on user request
+const setUserDataProcessingStatus = (
+  // tslint:disable-next-line: variable-name
+  _status: UserDataProcessingStatusEnum
+): TaskEither<Error, true> => notImplementedTask("setUserDataProcessingStatus");
+
+// after the operation, unblock the user to allow another login
+const post = (fiscalCode: FiscalCode): TaskEither<Error, true> =>
```
gunzip, line 110: I'd rather call these blockUser / unblockUser

GetUserDataProcessingActivity/handler.ts
```diff
@@ -0,0 +1,204 @@
...
+  userDataProcessingModel: UserDataProcessingModel
+) => (context: Context, input: unknown): Promise<ActivityResult> => {
+  /**
+   * Updates a UserDataProcessing record by creating a new version of it with a chenged status
+   * @param param0.currentRecord the record to be modified
+   * @param param0.nextStatus: the status to assign the record to
+   *
+   * @returns either an Error or the new created record
+   */
+  const fetchRecordFromDb = ({
```
gunzip, line 153: can you extract this method outside the body of `createGetUserDataProcessingActivityHandler` ?

gunzip: naming: `fetchRecordFromDb` ? -> `getUserDataProcessingRequest`

GetUserDataProcessingActivity/handler.ts
```diff
@@ -0,0 +1,204 @@
+/**
```
gunzip, line 1: Do we need this Activity ? I'm asking because the orchestrator will be triggered by an upcoming record inserted into user-data-requests collection so the input comes from the orchestrator client function bindings

gunzip: (it's ok to use a query in the client.ts)

balanza: I basically did it for not have to construct here models and require env keys (as we do already with other activities). I guess we can eventually remove it if we find it unhelpful in future refactoring

gunzip: yes we should convert this into an utility function only used by the cli (not by the orchestrator)

GetUserDataProcessingActivity/handler.ts
```diff
@@ -140,32 +140,34 @@ const logFailure = (context: Context) => (
   }
 };
 
+/**
```
gunzip, line 143: can you put this files inside the orchestrator / cli folder?

PR #55, opened 2020-05-21
[#172885142] backup and delete user data

balanza:
backup user data on a dedicated storage and then delete data from db

## Implemented models
* profile
* message
* message_status
* message_content
* notification
* notification_status

## usage
```sh
yarn build && node dist/UserDataDeleteOrchestrator/cli.js <FISCAL_CODE>
```


DeleteUserDataActivity/models/message_status.ts
```diff
@@ -0,0 +1,41 @@
...
+    messageId: NonEmptyString,
+    documentId: NonEmptyString
+  ): Promise<Either<DocumentDb.QueryError, string>> {
+    return DocumentDbUtils.deleteDocument(
+      this.dbClient,
+      this.collectionUri,
+      documentId,
+      messageId
+    );
+  }
```
gunzip, line 23: can't we just use `DocumentDbUtils.deleteDocument` where needed without extending the models ?

DeleteUserDataActivity/models/profile.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  /**
+   * Retrieves a list of every version of the requested model
+   * @param modelId
+   */
+  public findAllVersionsByModelId(
+    fiscalCode: FiscalCode
+  ): DocumentDbUtils.IResultIterator<RetrievedProfile> {
+    return DocumentDbUtils.findAllVersionsByModelId(
```
gunzip, line 31: can we use `DocumentDbUtils.findAllVersionsByModelId` without extending the model ?

DeleteUserDataActivity/utils/documentdb.ts
```diff
@@ -0,0 +1,109 @@
...
+          value: modelIdValue
+        },
+        {
+          name: "@partitionKey",
+          value: partitionKeyValue
+        }
+      ],
+      // do not use ${collectionName} here as it may contain special characters
+      query: `SELECT * FROM m WHERE (m.${modelIdField} = @modelId
+          AND m.${partitionKeyField} = @partitionKey) ORDER BY m.version DESC`
```
gunzip, line 55: why we need to order ?

DeleteUserDataActivity/utils/documentdb.ts
```diff
@@ -0,0 +1,109 @@
...
+): DocumentDbUtilsBase.IFoldableResultIterator<
+  Promise<ReadonlyArray<Either<DocumentDb.QueryError, string>>>
+> {
+  return DocumentDbUtilsBase.reduceResultIterator(
+    documentIterator,
+    (
+      prev: Promise<ReadonlyArray<Either<DocumentDb.QueryError, string>>>,
+      curr: T
+    ) =>
+      Promise.all([prev, deleteDocument(client, collectionUri, curr.id)]).then(
```
gunzip, line 74:
```suggestion
      prev.then(() => deleteDocument(client, collectionUri, curr.id)).then(
```

gunzip: maybe I'm wrong (I'm taking a quick look) but if you do it like that you'll trigger two promises in parallel for each run, the next iteration you will call Promise.all() on the same promises and son on...

GetUserDataProcessingActivity/handler.ts
```diff
@@ -0,0 +1,212 @@
...
+    case "RECORD_NOT_FOUND":
+      context.log.error(`${logPrefix}|Error record not found |ERROR=`);
+      break;
+    default:
+      assertNever(failure);
+  }
+};
+
+/**
+ * Updates a UserDataProcessing record by creating a new version of it with a chenged status
```
gunzip, line 144: it looks like it does not update..

GetUserDataProcessingActivity/handler.ts
```diff
@@ -0,0 +1,212 @@
+/**
+ * Updates the status of a UserDataProcessing record
```
gunzip, line 2: updates ?

GetUserDataProcessingActivity/handler.ts
```diff
@@ -0,0 +1,212 @@
+/**
```
gunzip, line 1: why do we need this activity ? it looks like it's a convenience method used only by the cli

DeleteUserDataActivity/utils/documentdb.ts
balanza: The difference is: in the first case operations are executed independently, but results are collected sequentially; in the latter operations are executed sequentially.

balanza:
@gunzip marked as draft as it has two unused functions declared (and consequently lint errors).
Need to close it tomorrow

DeleteUserDataActivity/handler.ts
```diff
@@ -0,0 +1,617 @@
...
+
+/**
+ * to cast an error to QueryFailure
+ * @param err
+ */
+const toQueryFailure = (err: Error | QueryError): QueryFailure =>
+  QueryFailure.encode({
+    kind: "QUERY_FAILURE",
+    reason:
+      err instanceof Error ? err.message : `QueryError: ${JSON.stringify(err)}`
```
gunzip, line 147: consider using toError (from Either module)

DeleteUserDataActivity/handler.ts
```diff
@@ -0,0 +1,617 @@
...
+ * @param data serializable data to be saved
+ *
+ * @returns either a blob failure or the saved object
+ */
+export const saveDataToBlob = <T>(
+  { blobService, containerName, folder }: IBlobServiceInfo,
+  blobName: string,
+  data: T
+): TaskEither<BlobCreationFailure, T> =>
+  taskify(blobService.createBlockBlobFromText)(
```
gunzip, line 217:
```suggestion
  taskify(blobService.createBlockBlobFromText.bind(blobService)(
```

we don't know if `createBlockBlobFromText` is using a referente to '_this_' (probably not, but since we cannot know we have to bind the method anyway)

DeleteUserDataActivity/handler.ts
```diff
@@ -0,0 +1,617 @@
...
+}: {
+  messageModel: MessageModel;
+  userDataBackup: IBlobServiceInfo;
+  message: RetrievedMessageWithoutContent;
+}): TaskEither<
+  BlobCreationFailure | QueryFailure | DocumentDeleteFailure,
+  RetrievedMessageWithoutContent
+> => {
+  return sequenceT(taskEitherSeq)<
+    QueryFailure | BlobCreationFailure | DocumentDeleteFailure,
```
gunzip, line 442: maybe it's worth to define a type alias for `QueryFailure | BlobCreationFailure | DocumentDeleteFailure`

gunzip:
is the description of this pr still relevant ? what's the command to run and what it does ?

balanza:
This PR is on hold because we need to figure out a better way to interact with Redis (this service should not have access to the redis node)

balanza:
This PR will be replaced by https://github.com/pagopa/io-functions-admin/pull/62

However, in this branch is includes a CLI script that may server for backoffice and troubleshooting.

Consider merge this as well (after rebase against #62) to include such functionality

PR #58, opened 2020-07-07
[#172884826] automatic user data download

balanza:
This PR implements the trigger that automatically starts the procedure that will allow user to download their data, when requested.

### Implementation
* a function is listening to changes in the collection of users' data processing requests
* documents are filtered and, for each of those which must be processed, an orchestrator is started 
* ~each orchestrator is paused 24h as in specifications, then resumes the procedure~ no more


UserDataDownloadSubOrchestrator/handler.ts
```diff
@@ -0,0 +1,161 @@
...
+    kind: "ACTIVITY",
+    reason: readableReport(err)
+  });
+
+export const handler = function*(
+  context: IFunctionContext,
+  document: unknown
+): IterableIterator<unknown> {
+  // This check has been done on the parent orchestrator, so it should never fail.
+  // However, it's maybe woth the effort of check it twice
```
gunzip, line 76:
```suggestion
  // However, it's worth the effort to check it twice
```

UserDataDownloadSubOrchestrator/handler.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  if (isLeft(invalidInputOrCurrentUserDataProcessing)) {
+    return invalidInputOrCurrentUserDataProcessing.value;
+  }
+
+  const currentUserDataProcessing =
+    invalidInputOrCurrentUserDataProcessing.value;
+
+  // start this operation tomorrow
+  yield context.df.createTimer(new Date(Date.now() + 24 * 60 * 60 * 100));
```
gunzip, line 99: you cannot use Date.now() inside an orchestrator, you should use context.df.currentUtcDateTime instead, see

https://github.com/pagopa/io-functions-bonus/blob/master/EligibilityCheckOrchestrator/index.ts#L229

and https://docs.microsoft.com/it-it/azure/azure-functions/durable/durable-functions-code-constraints

UserDataDownloadOrchestrator/__tests__/handler.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      return value;
+    }
+    prevValue = value;
+  }
+};
+
+// just a convenient cast, good for every test case
+const context = (mockOrchestratorContext as unknown) as IFunctionContext;
+
+describe("handler", () => {
```
gunzip, line 52:
```suggestion
describe(" UserDataDownloadOrchestrator", () => {
```

UserDataDownloadSubOrchestrator/__tests__/handler.test.ts
```diff
@@ -0,0 +1,306 @@
...
+  }
+};
+
+// just a convenient cast, good for every test case
+const context = (mockOrchestratorContext as unknown) as IFunctionContext;
+
+const handler = getHandler(DELAY);
+
+// tslint:disable-next-line: no-big-function
+describe("handler", () => {
```
gunzip, line 88:
```suggestion
describe(" UserDataDownloadSubOrchestrator", () => {
```

UserDataDownloadSubOrchestrator/handler.ts
```diff
@@ -0,0 +1,169 @@
...
+      const bundle = ExtractUserDataActivityResultSuccess.decode(
+        yield context.df.callActivity("ExtractUserDataActivity", {
+          fiscalCode: currentUserDataProcessing.fiscalCode
+        })
+      ).getOrElseL(err => {
+        throw toActivityFailure(err, "ExtractUserDataActivity");
+      });
+
+      SendUserDataDownloadMessageActivityResultSuccess.decode(
+        yield context.df.callActivity("SendUserDataDownloadMessageActivity", {
```
gunzip, line 125: withRetry ?

UserDataDownloadSubOrchestrator/index.ts
```diff
@@ -0,0 +1,15 @@
+import * as df from "durable-functions";
+import { Hour, Millisecond } from "italia-ts-commons/lib/units";
+import { getHandler } from "./handler";
+
+const delayInHours = (typeof process.env.USER_DATA_DOWNLOAD_DELAY_HOURS ===
```
gunzip, line 5: `NonNegativeInteger.decode(process.env.USER_DATA_DOWNLOAD_DELAY_HOURS)
   .getOrElse(24 as NonNegativeInteger)` 

?

UserDataDownloadSubOrchestrator/index.ts
```diff
@@ -0,0 +1,15 @@
...
+"undefined"
+  ? 24
+  : process.env.USER_DATA_DOWNLOAD_DELAY_HOURS) as Hour;
+const delay = (delayInHours * 60 * 60 * 1000) as Millisecond;
```
gunzip, line 9:
```suggestion
const delayMs = (delayInHours * 60 * 60 * 1000) as Millisecond;
```

UserDataDownloadSubOrchestrator/handler.ts
```diff
@@ -0,0 +1,169 @@
...
+  extra?: object
+) =>
+  ActivityFailure.encode({
+    activityName,
+    extra,
+    kind: "ACTIVITY",
+    reason: readableReport(err)
+  });
+
+export const getHandler = (delay: Millisecond = 0 as Millisecond) =>
```
gunzip, line 72:
```suggestion
export const getHandler = (delayMs: Millisecond = 0 as Millisecond) =>
```

UserDataDownloadOrchestratorTrigger/index.ts
```diff
@@ -0,0 +1,7 @@
+import { Context } from "@azure/functions";
+import * as df from "durable-functions";
+
+export default (context: Context, input: unknown) =>
+  df
```
gunzip, line 5: what about moving the code which is currently into  `UserDataDownloadOrchestrator` here and
call `startNew` with  `UserDataDownload(Sub)Orchestrator` ? To avoid a layer of indirection

balanza: It's actually a good idea, and it will help us when we'll have to handle DELETE requests too.

balanza:
@gunzip changes introduced with latest commits (besided your feedbacks):
* `UserDataDownloadOrchestratorTrigger` is now `UserDataProcessingTrigger` and includes the filtering logic that was on (previous) `UserDataDownloadOrchestrator`
* (previous) `UserDataDownloadOrchestrator` has been deleted
* `UserDataDownloadSubOrchestrator` is now `UserDataDownloadOrchestrator`

UserDataProcessingTrigger/index.ts
```diff
@@ -0,0 +1,69 @@
...
+    );
+
+  context.log.info(
+    `${logPrefix}: processing ${tasksDescriptors.length} document${
+      tasksDescriptors.length === 1 ? "" : "s"
+    }`
+  );
+
+  tasksDescriptors.forEach(task =>
+    df.getClient(context).startNew(task.orchestrator, undefined, task.input)
```
gunzip, line 67: i think that we need to await Promise.all here otherwise the process will end before promises are settled

UserDataProcessingTrigger/index.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  context.log.info(
+    `${logPrefix}: processing ${tasksDescriptors.length} document${
+      tasksDescriptors.length === 1 ? "" : "s"
+    }`
+  );
+
+  const startAllNew = () =>
+    tasksDescriptors.map(({ orchestrator, input: orchestratorInput }) =>
+      dfClient.startNew(orchestrator, undefined, orchestratorInput)
```
gunzip, line 69: can we use the userProcessingDataId (cf + choice) as instanceid for the orchestartor ? this would simplify troubleshooting

README.md
```diff
@@ -26,10 +26,8 @@ they may be customized as needed.
 | AZURE_APIM_HOST             | The host name of the API Management service                                                      | string |
 | AZURE_APIM_RESOURCE_GROUP   | The name of the resource group used to get the subscriptions                                     | string |
 | AZURE_SUBSCRIPTION_ID       | Credentials which identify the Azure subscription, used to init the APIM  client                 | string |
-| UserDataArchiveStorageConnection | Storage connection string to store zip file for user to download their data | string |
```
balanza: @gunzip it's still used here: https://github.com/pagopa/io-functions-admin/blob/e50ba7fe3ef4c7c47dc39471afafb14eb3c35ad2/ExtractUserDataActivity/index.ts#L78

We should change it as well. Should we use `StorageConnection`? I guess we meant to keep user data bundles separated from normal IO's working storage

gunzip: no sorry it's my fault, I'll revert it into the readme again :)

PR #59, opened 2020-07-09
[#171683039] - Porting of municipalities exports from io-service-metadata

AleDore:
Municipalities porting and refactor


utils/municipality.ts
```diff
@@ -0,0 +1,339 @@
...
+  municipalityToCatastale: Map<string, string>
+): TaskEither<Error, ReadonlyArray<ISerializableMunicipality>> =>
+  getFileFromBlob(
+    blobService,
+    MUNICIPALITIES_CONTAINER_NAME,
+    ABOLISHED_MUNICIPALITIES_BLOB_NAME
+  )
+    .chain(rawFile =>
+      rawFile.foldL(
+        () => fromEither(leftE(new Error("abolished municipalities is empty"))),
```
gunzip, line 217: fromEither(left) === fromLeft()

utils/municipality.ts
```diff
@@ -0,0 +1,339 @@
...
+  )
+    .foldTaskEither<Error, StringMatrix>(
+      err => fromLeft(err),
+      _ =>
+        _.foldL(
+          () => fromLeft(new Error("Municipalities with catastale is empty")),
+          __ =>
+            parseCsv(__, optionMunicipalitiesWithCatastale).foldTaskEither(
+              e => fromEither(leftE(e)),
+              matrixString => fromEither(rightE(matrixString))
```
balanza, line 164:
```suggestion
              matrixString => taskEither.of(matrixString)
```

utils/municipality.ts
```diff
@@ -0,0 +1,339 @@
...
+    MUNICIPALITIES_CATASTALI_BLOB_NAME
+  )
+    .foldTaskEither<Error, StringMatrix>(
+      err => fromLeft(err),
+      _ =>
+        _.foldL(
+          () => fromLeft(new Error("Municipalities with catastale is empty")),
+          __ =>
+            parseCsv(__, optionMunicipalitiesWithCatastale).foldTaskEither(
+              e => fromEither(leftE(e)),
```
balanza, line 163:
```suggestion
	        e => fromLeft(e),
```

utils/municipality.ts
```diff
@@ -0,0 +1,339 @@
...
+          fromLeft<Error, StringMatrix>(
+            new Error("Elenco codici e denominazioni is empty")
+          ),
+        csv => parseCsv(csv, optionCsvParseForeignCountries)
+      )
+    )
+    .chain(result =>
+      array.sequence(taskEither)(
+        result
+          .filter(r => r[9] !== "")
```
gunzip, line 322: a comment here on what's happening would be useful or better you can isolate this filter into its own method with a meaningful name

utils/municipality.ts
```diff
@@ -0,0 +1,339 @@
...
+            decodeForeignCountry(r).fold(
+              e =>
+                fromLeft(
+                  new Error(
+                    `Cannot decode foreign municipality| ${e.toString()}`
+                  )
+                ),
+              municipality =>
+                serializeMunicipalityToJson(blobService, {
+                  codiceCatastale: r[9],
```
gunzip, line 333: maybe an object with a map between indexex (integers) and column names (constants) can be useful here

utils/municipality.ts
```diff
@@ -0,0 +1,339 @@
...
+            municipalitiesCatastaleRows.reduce(
+              (map: Map<string, string>, row) => {
+                map.set(row[1].toLowerCase(), row[0]);
+                return map;
+              },
+              new Map<string, string>()
+            )
+          )
+        )
+    );
```
balanza, line 182: Is `foldTaskEither` needed? as you are not lifting left type, I think you can use `map`

```typescript
    .map(
      municipalitiesCatastaleRows =>
          municipalitiesCatastaleRows.reduce(
            (map: Map<string, string>, row) => {
              map.set(row[1].toLowerCase(), row[0]);
              return map;
            },
            new Map<string, string>()
          )
       );
```
(to be verified)

utils/municipality.ts
```diff
@@ -0,0 +1,339 @@
...
+
+const fromAbolishedMunicipalityToSerializableMunicipality = (
+  abolishedMunicipality: t.TypeOf<typeof AbolishedMunicipality>,
```
balanza, line 185:
```suggestion
  abolishedMunicipality: AbolishedMunicipality,
```

utils/municipality.ts
```diff
@@ -0,0 +1,339 @@
...
+    municipality: {
+      codiceProvincia: "",
+      codiceRegione: "",
+      denominazione: abolishedMunicipality.comune,
+      denominazioneInItaliano: abolishedMunicipality.comune,
+      denominazioneRegione: "",
+      siglaProvincia: abolishedMunicipality.provincia
+    }
+  } as ISerializableMunicipality;
+};
```
balanza, line 199: I'm afraid of this cast. The object isn't recognised as `ISerializableMunicipality` because we're not sure `municipality` is a `Municipality`.

Consider decode the municipality field value:

```typescript
   municipality: Municipality.decode({
      codiceProvincia: "",
      codiceRegione: "",
      denominazione: abolishedMunicipality.comune,
      denominazioneInItaliano: abolishedMunicipality.comune,
      denominazioneRegione: "",
      siglaProvincia: abolishedMunicipality.provincia
    }).getOrElseL(_ => { throw new Error(/* err message */); }
```

Or even, why not make `ISerializableMunicipality` a decoder itself?

utils/municipality.ts
balanza, line 217:
```suggestion
        () => fromLeft(new Error("abolished municipalities is empty")),
```

utils/municipality.ts
```diff
@@ -0,0 +1,339 @@
...
+): TaskEither<Error, ReadonlyArray<ISerializableMunicipality>> =>
+  getFileFromBlob(
+    blobService,
+    MUNICIPALITIES_CONTAINER_NAME,
+    ABOLISHED_MUNICIPALITIES_BLOB_NAME
+  )
+    .chain(rawFile =>
+      rawFile.foldL(
+        () => fromEither(leftE(new Error("abolished municipalities is empty"))),
+        json => fromEither(rightE(JSON.parse(json)))
```
balanza, line 218:
```suggestion
        json => taskEither.of(JSON.parse(json)))
```

Besides of that, be careful because `JSON.parse` can throw! Consider use [parseJSON](https://github.com/gcanti/fp-ts/blob/1.x/test/Either.ts#L533)

utils/municipality.ts
AleDore: I changed it into a `chain`

utils/file.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export const getFileFromBlob = (
+  blobService: BlobService,
+  containerName: NonEmptyString,
+  blobName: NonEmptyString
+): TaskEither<Error, Option<string>> =>
+  tryCatch(
+    () => getBlobAsText(blobService, containerName, blobName).then(e => e),
+    toError
+  ).foldTaskEither(
```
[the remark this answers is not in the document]
AleDore: `getBlobAsText` returns `Promise<Either<Error, BlobResult>>` so I've used the same code to turn it into a TaskEither -> see [https://github.com/pagopa/io-functions-admin/blob/cdae1896d9ff94180e4b687a226508a800ae1110/ExtractUserDataActivity/handler.ts#L161](https://github.com/pagopa/io-functions-admin/blob/cdae1896d9ff94180e4b687a226508a800ae1110/ExtractUserDataActivity/handler.ts#L161)

utils/file.ts
```diff
@@ -0,0 +1,81 @@
...
+    () =>
+      upsertBlobFromText(
+        blobService,
+        containerName,
+        blobName,
+        textContent
+      ).then(e => e),
+    toError
+  ).foldTaskEither(
+    err => fromEither(left(err)),
```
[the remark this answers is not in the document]
AleDore: same as previous

utils/municipality.ts
```diff
@@ -0,0 +1,339 @@
...
+  record: ReadonlyArray<string>
+): t.Validation<Municipality> => {
+  if (record.length < 13) {
+    return leftE([
+      {
+        context: [],
+        value: "record has not the right length"
+      }
+    ]);
+  }
```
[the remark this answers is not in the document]
AleDore: I think that this check is trying to avoid column shifting problems, because if a record length is larger than expected, we can assign a wrong value to a property. Moreover we use a positional notation to extract values from csv files.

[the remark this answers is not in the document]
balanza: > if a record length is larger than expected

Actually the check is for records **shorter** than expected, isn't it?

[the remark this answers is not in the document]
AleDore: Yes, but the logic is the same, maybe we can turn this check into `record.length !== <expected_record_length>`

UpdateMunicipalities/function.json
```diff
@@ -0,0 +1,11 @@
+{
+  "bindings": [
+    {
+      "schedule": "0 */15 * * * *",
```
gunzip, line 4: how many times is this supposed to run ? once a day suffices

UpdateMunicipalities/handler.ts
```diff
@@ -0,0 +1,31 @@
...
+  blobService: BlobService
+) => async (context: Context): Promise<unknown> => {
+  return sequenceT(taskEither)(
+    exportAbolishedMunicipality(blobService),
+    exportCurrentMunicipalities(blobService),
+    exportForeignMunicipalities(blobService)
+  )
+    .foldTaskEither<Error, void>(
+      err => {
+        context.log.error(`Cannot update municipalities| ${err}`);
```
gunzip, line 25: may we prefix the error with the name of the function? moreover get rid of the space after '|'

UpdateMunicipalities/handler.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export const getUpdateMunicipalitiesHandler = (
+  blobService: BlobService
+) => async (context: Context): Promise<unknown> => {
+  return sequenceT(taskEither)(
+    exportAbolishedMunicipality(blobService),
+    exportCurrentMunicipalities(blobService),
+    exportForeignMunicipalities(blobService)
+  )
+    .foldTaskEither<Error, void>(
```
gunzip, line 23: it looks like `mapLeft()` would suffice here (instead of folding)

UpdateMunicipalities/index.ts
```diff
@@ -0,0 +1,12 @@
+﻿import { createBlobService } from "azure-storage";
+
+import { getRequiredStringEnv } from "io-functions-commons/dist/src/utils/env";
+
+import { getUpdateMunicipalitiesHandler } from "./handler";
+
+const storageConnectionString = getRequiredStringEnv("StorageConnection");
```
gunzip, line 7:
```suggestion
const storageConnectionString = getRequiredStringEnv("AssetsStorageConnection");
```

utils/municipality.ts
```diff
@@ -0,0 +1,365 @@
...
+    codiceCatastale,
+    municipality: {
+      codiceProvincia: "",
+      codiceRegione: "",
+      denominazione: abolishedMunicipality.comune,
+      denominazioneInItaliano: abolishedMunicipality.comune,
+      denominazioneRegione: "",
+      siglaProvincia: abolishedMunicipality.provincia
+    }
+  }).getOrElseL(_ => {
```
gunzip, line 216: may we just return the either and skip all the invalid entries with a warning?

utils/municipality.ts
```diff
@@ -0,0 +1,365 @@
...
+            )
+          )
+      )
+    )
+    .chain((abolishedMunArray: AbolishedMunicipalityArray) =>
+      taskEither.of(
+        abolishedMunArray
+          .filter(am => municipalityToCatastale.has(am.comune.toLowerCase()))
+          .map(am =>
+            fromAbolishedMunicipalityToSerializableMunicipality(
```
gunzip, line 254: this may throw currently, see comment above

utils/municipality.ts
```diff
@@ -0,0 +1,365 @@
...
+        () =>
+          fromLeft<Error, StringMatrix>(
+            new Error("Elenco codici e denominazioni is empty")
+          ),
+        csv => parseCsv(csv, optionCsvParseForeignCountries)
+      )
+    )
+    .chain(result =>
+      array.sequence(taskEither)(
+        filterEmptyCodiceCatastaleRecords(result, 9 as NonNegativeInteger).map(
```
gunzip, line 346: why "9" ? I guess the index value must be hardcoded into `filterEmptyCodiceCatastaleRecords` instead of being taken as paramter

gunzip:
why draft?=

utils/municipality.ts
AleDore: agree, just refactor

UpdateMunicipalities/function.json
AleDore: Which hour?

AleDore: I've set every day at 01:00 AM

UpdateMunicipalities/handler.ts
```diff
@@ -17,14 +21,45 @@ const logPrefix = "getUpdateMunicipalitiesHandler";
 export const getUpdateMunicipalitiesHandler = (
   blobService: BlobService
 ) => async (context: Context): Promise<unknown> => {
-  return sequenceT(taskEither)(
+  const partialResults = sequenceT(taskEitherSeq)(
```
AleDore, line 24: I've to refactor this way (but it's not the only way I think), in order to avoid Task maximum number exceeded. All feedbacks are welcome!

balanza:
@AleDore I assume this to not be actual anymore

AleDore:
Yes it is correct!

Inviato da iPhone

PR #62, opened 2020-07-16
[#172884867] Automatic user data delete on GDPR right to erasure claim

balanza:
This PR introduces into this service the ability to autonomously start the job that backup and deletes user data on user's request.

~It depends on this to be merged: https://github.com/pagopa/io-backend/pull/691~

### Done
* trigger on `user-data-processing` collection
* orchestrator workflow with user's session lock/unlock
* api interaction with Session Api endpoints for locking/unlocking
* start delete data activity
* wait for 7 days before effectively delete data
* handle request cancelling
* make wait time an env parameter (for testing purpose)

### Things worth looking
* How we handle abortions in the [orchestrator](https://github.com/pagopa/io-functions-admin/pull/62/files#diff-922e66ef5220937f700aa6f582ba5471R222) and in the [trigger](https://github.com/pagopa/io-functions-admin/pull/62/files#diff-cc571ad4d7d64e8985b2809080593f2fR97)
* How we care about concurrent pending/wip download requests in the [orchestrator](https://github.com/pagopa/io-functions-admin/pull/62/files#diff-922e66ef5220937f700aa6f582ba5471R222) 

### Env

| Variable name                    | Description                                                                                      | type   |
|----------------------------------|--------------------------------------------------------------------------------------------------|--------|
| SESSION_API_URL                  | Internal URL of the BACKEND API used to handle session lock/unlock requests                      | string |
| SESSION_API_KEY                  | service access key for the session API                                                           | string |
| USER_DATA_BACKUP_CONTAINER_NAME  | Name of the storage container in which user data is backuped before being permanently deleted    | string |
| USER_DATA_DELETE_DELAY_DAYS      | How many days to wait when a user asks for cancellation before effectively delete her data       | number |
| UserDataBackupStorageConnection  | Storage connection string for GDPR user data storage                                             | string |

### Feature flags

| Variable name                | Description                                       | default |
|------------------------------|---------------------------------------------------|---------|
| FF_ENABLE_USER_DATA_DOWNLOAD | Users' GDPR data access claims are processed      | true    |
| FF_ENABLE_USER_DATA_DELETE   | Users' GDPR right to erasure claims are processed | true    |

### Notable Changes
* introduced a `utils/fetch` module with pre-built timeout fetch api.
* introduced a `utils/appinsightsEvens` module with Application Insights helpers.
* introduced `XyzDeletableModel`s that extend `Xyz`s models by adding deleting operations (example: `ProfileDeletableModel` extends `ProfileModel`). Such modules are located in `utils/extensions`.


SetUserSessionLockActivity/handler.ts
```diff
@@ -0,0 +1,176 @@
...
+        return failure;
+      },
+      _ => {
+        context.log.info(`${logPrefix}|INFO|Activity succeeded`);
+        return ActivityResultSuccess.encode({ kind: "SUCCESS" });
+      }
+    )
+
+    .run()
+    .then(e => e.value);
```
gunzip, line 176: you can use fold above instead of bimap and remove then()

SetUserSessionLockActivity/handler.ts
```diff
@@ -0,0 +1,176 @@
...
+      )
+    )
+    .chain(({ action, fiscalCode }) =>
+      callSessionApi(sessionApiClient, action, fiscalCode)
+    )
+    .bimap(
+      failure => {
+        context.log.error(`${logPrefix}|ERROR|Activity failed`, failure);
+
+        // if its a transient failure we let the activity throw, so the orchestrator can retry
```
gunzip, line 163:
```suggestion
        // in case of transient failures we let the activity throw, so the orchestrator can retry
```

SetUserSessionLockActivity/index.ts
```diff
@@ -0,0 +1,18 @@
...
+import nodeFetch from "node-fetch";
+import { Client, createClient } from "../utils/sessionApiClient";
+import { createSetUserSessionLockActivityHandler } from "./handler";
+
+const sessionApiUrl = getRequiredStringEnv("SESSION_API_URL");
+const sessionApiKey = getRequiredStringEnv("SESSION_API_KEY");
+
+const client: Client<"ApiKey"> = createClient({
+  baseUrl: sessionApiUrl,
+  fetchApi: (nodeFetch as unknown) as typeof fetch,
```
gunzip, line 11: we must use a fetch with timeout like in https://github.com/pagopa/io-functions-admin/blob/master/SendUserDataDownloadMessageActivity/index.ts#L25

SetUserSessionLockActivity/handler.ts
```diff
@@ -0,0 +1,176 @@
...
+ * Wraps the logic to call Session API and lift errors to the correct domain values
+ * @param sessionApiClient
+ * @param action
+ * @param fiscalCode
+ */
+const callSessionApi = (
+  sessionApiClient: Client<"ApiKey">,
+  action: ActivityInput["action"],
+  fiscalCode: FiscalCode
+): TaskEither<ApiCallFailure | BadApiRequestFailure, SuccessResponse> =>
```
AleDore, line 88: Can we add a super type for `ApiCallFailure` and `BadApiRequestFailure`

UserDataDeleteOrchestrator/handler.ts
```diff
@@ -0,0 +1,184 @@
...
+      })
+    ).getOrElseL(err => {
+      throw toActivityFailure(err, "SetUserDataProcessingStatusActivity", {
+        status: UserDataProcessingStatusEnum.CLOSED
+      });
+    });
+    return OrchestratorSuccess.encode({ kind: "SUCCESS" });
+  } catch (error) {
+    context.log.error(
+      `${logPrefix}|ERROR|Failed processing user data for download: ${error.message}`
```
gunzip, line 162: error is any here (not an instance of Error) so may not have a `message` property

package.json
```diff
@@ -3,19 +3,21 @@
   "version": "0.5.0",
   "license": "MIT",
   "scripts": {
-    "build": "yarn generate:definitions && tsc",
+    "prebuild": "npm-run-all --parallel generate:*",
```
gunzip, line 6: will need rebase after https://github.com/pagopa/io-functions-admin/pull/63 is merged

utils/sessionApiClient.ts
```diff
@@ -0,0 +1,167 @@
+/**
+ * Do not edit this file it is auto-generated by italia-utils / gen-api-models.
```
gunzip, line 2: if it is auto-generated it shoulnd't be committed (maybe the comment is outdated?)

SetUserSessionLockActivity/handler.ts
balanza: We can union such in a type that models "failures that can happen when calling api".
I think it would only make this function more obscure. 
I'd rather use `ActivityResultFailure` to indicate "hey, don't care about the kind of failure that can happen here".

utils/sessionApiClient.ts
balanza: Not actually. It's generated by an external script (newer version of `io-utils`). I thought it'd be valuable to remind that this is generated code

balanza:
This is kinda final form for this PR. It must be integrated with intensive unit tests, but it's somehow definitive now.

balanza:
This PR is huge. I tried to sum up relevant changes in the PR description. Please have a look there before dig into code.

UserDataProcessingTrigger/index.ts
```diff
@@ -22,57 +29,100 @@ export const ProcessableUserDataDownload = t.intersection([
...
+                )}]`
+              );
+              return lazyOperations;
+            },
+            lazyOp => [...lazyOperations, lazyOp]
+          ),
+      [] as ReadonlyArray<Lazy<Promise<string | void>>>
     );
 
   context.log.info(
```
gunzip, line 121: instead of logging here, can we add a log in the switch for each case? (so we can track the operation type)

balanza: should we use AI events?

gunzip: yes that would be great

DeleteUserDataActivity/backupAndDelete.ts
```diff
@@ -0,0 +1,472 @@
...
+const backupAndDeleteNotification = ({
+  notificationModel,
+  userDataBackup,
+  notification
+}: {
+  notificationModel: NotificationDeletableModel;
+  userDataBackup: IBlobServiceInfo;
+  notification: RetrievedNotification;
+}): TaskEither<DataFailure, RetrievedNotification> => {
+  return sequenceT(taskEitherSeq)<
```
cloudify, line 150: In general if you have a single return statement, just avoid the code block

DeleteUserDataActivity/backupAndDelete.ts
```diff
@@ -0,0 +1,472 @@
...
+} from "io-functions-commons/dist/src/utils/documentdb";
+
+/**
+ * Recursively consumes an iterator and executes operations on every item
+ * @param deleteSingle takes an item and delete it
+ * @param userDataBackup references about where to save data
+ * @param makeBackupBlobName takes an item and construct a name for the backup blob
+ * @param iterator an iterator of every result from the db
+ */
+const executeRecursiveBackupAndDelete = <T>(
```
cloudify, line 50: It would be nice to generalize the recoursive logic and separate it from the specific use case (I.e the accumulators)

DeleteUserDataActivity/backupAndDelete.ts
```diff
@@ -0,0 +1,472 @@
...
+                  saveDataToBlob<T>(
+                    userDataBackup,
+                    makeBackupBlobName(item),
+                    item
+                  ),
+                  tryCatch(() => deleteSingle(item), toError)
+                    .mapLeft(toDocumentDeleteFailure)
+                    .chain(_ => fromEither(_).mapLeft(toDocumentDeleteFailure)),
+                  // recursive step
+                  executeRecursiveBackupAndDelete<T>(
```
cloudify, line 94: Is this basically a reduce operation on the iteration?

DeleteUserDataActivity/backupAndDelete.ts
```diff
@@ -0,0 +1,472 @@
...
+    )
+      // type lift
+      // from TaskEither of Either of Option of X
+      // to TaskEither of X
+      // this way we collaps every left/none case into the same path
+      .chain(fromEither)
+      .chain(e => fromEither(fromOption(undefined)(e)))
+      .foldTaskEither<DataFailure, Option<MessageContent>>(
+        _ => {
+          // unfortunately, a document not found is threated like a query error
```
cloudify, line 260: Does this mean that the user may get a subset of his data￼?

balanza: Yes, in case of an error while fetching data. The rationale is: we don't want to impede the operation in case of a missing blob (which is a case of inconsistence of data, and should never happen).

DeleteUserDataActivity/backupAndDelete.ts
balanza: It's basically [this](https://github.com/pagopa/io-functions-commons/blob/b99194e201b015e4bce9c4dec7692afadc6d8d9a/src/utils/documentdb.ts#L544), but I didn't get to fit it in at the time of writing

balanza:
> what's the expected string value to obtain 'false' ?

Anything different from `1`

PR #69, opened 2020-09-04
[#174644239] - Fix GetServices returns all services

AleDore:
This PR includes a fix to GetServices that now returns all services, instead of a subset.
This also includes specs update on GetServices return type.

### DIFF
--- a/GetServices/handler.ts
+++ b/GetServices/handler.ts
@@ -32,11 +32,15 @@ import {
   ResponseSuccessJson
 } from "italia-ts-commons/lib/responses";
 import { Service as ApiService } from "../generated/definitions/Service";
+import { ServiceIdWithVersion } from "../generated/definitions/ServiceIdWithVersion";
 import { retrievedServiceToApiService } from "../utils/conversions";
 
 type IGetServicesHandlerResult =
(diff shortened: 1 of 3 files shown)


GetServices/handler.ts
```diff
@@ -69,13 +76,23 @@ export function GetServicesHandler(
...
         error => ResponseErrorQuery("Cannot get services", error),
-        iteratorResults => {
+        results => {
+          // tslint:disable-next-line: no-inferred-empty-object-type
+          const reducedResults = results.reduce((prev, curr) => {
+            return {
+              ...prev,
+              ...curr
+            };
+          }, {});
```
gunzip, line 92: what's the purpose of this code?

AleDore: results' type is `{}[]`. I've used reduce to flatten results into `{}`

AleDore: nevermind, changed implementation in [85a8305](https://github.com/pagopa/io-functions-admin/pull/69/commits/85a8305ddacd25efb495fa696662e5479d9f7740)

gunzip (approved):

PR #72, opened 2020-09-11
[#174710415] Send confirmation email to user after data delete

balanza:
Send a confirmation email to a user after his/her data is been deleted. 

### QUESTIONS
* ~Is it correct to send email to every user, regarding the fact that the email si validated or not?~
* ~How to deal the case in which the user has no email?~


GetProfileActivity/handler.ts
```diff
@@ -0,0 +1,156 @@
+/**
+ * Updates the status of a Profile record
```
gunzip, line 2: needs to be updated

GetProfileActivity/handler.ts
```diff
@@ -0,0 +1,156 @@
...
+    .map(record =>
+      ActivityResultSuccess.encode({
+        kind: "SUCCESS",
+        value: record
+      })
+    )
+    .mapLeft(failure => {
+      logFailure(context)(failure);
+      return failure;
+    })
```
gunzip, line 153: you can use fold and avoid then()

gunzip (commented):
I agree that there's a lot of duplicated code that must be avoid (to send emails), it's probably better to resolve this issue then return on this again

balanza:
closed and reopened at https://github.com/pagopa/io-functions-admin/pull/86

PR #81, opened 2020-09-17
[#174824788] subscription feed on user delete

balanza:
Add a record of type `UNSUBSCRIBE` to the subscription feed when a user is deleted in his/her claim to erasure.

A connection to the SubscriptionsFeed storage table is added to the current service.

### Env variables
The following environment variables have been added:

| Variable name                    | Description                                                                                      | type   |
|----------------------------------|--------------------------------------------------------------------------------------------------|--------|
| QueueStorageConnection           | Storage connection string for subscription feed                                                  | string |
| SUBSCRIPTIONS_FEED_TABLE         | Table name for the Subscriptions Feed in the storage                                             | string |


UpdateSubscriptionsFeedActivity/index.ts
```diff
@@ -0,0 +1,215 @@
...
+      subscriptionKind: t.literal("SERVICE")
+    })
+  ])
+]);
+
+export type Input = t.TypeOf<typeof Input>;
+
+// When the function starts, attempt to create the table if it does not exist
+// Note that we cannot log anything just yet since we don't have a Context
+tableService.createTableIfNotExists(subscriptionsFeedTable, () => 0);
```
gunzip, line 61: remove these lines (they're useless even in io-fn-app)

env.example
```diff
@@ -53,3 +53,6 @@ NODE_TLS_REJECT_UNAUTHORIZED=0
 languageWorkers__node__arguments="--inspect=0.0.0.0:5861"
 
 # AzureWebJobs.UpdateVisibleServicesCache.Disabled=true
+
+QueueStorageConnection=<STORAGE_CONNECTION_STRING>
```
gunzip, line 57: can we rename `QueueStorageConnection` to `SubscriptionFeedStorageConnection` ?

gunzip (changes requested):

env.example
balanza: Of course, just note that it's the same value used here https://github.com/pagopa/io-functions-app/blob/master/UpdateSubscriptionsFeedActivity/index.ts#L18

gunzip (approved):
lgtm @francescopersico can we merge this?

UpdateSubscriptionsFeedActivity/index.ts
```diff
@@ -0,0 +1,176 @@
+﻿import * as crypto from "crypto";
```
francescopersico, line 1: Do we really want to maintain all this logic when we just want to add an UNSUBSCRIBE event in the table?

gunzip: May you mark the parts that you think are redundant? I've asked the same question myself, then I noticed that the intent here was to have a general purpose activity (as the name is indeed 'UpdateSubscriptionsFeedActivity' and not 'UnsubscribeUser'). If we switch this to `UnsubscribeUser` we still need to keep 80% of the logic and you only gain about 10 lines less getting rid of the 'subscribe' branch.

francescopersico: The risk of duplicating the "subscribe" code here is that you must maintain it in sync if changes happens.
About sharing the code inside io-functions-commons i am more convinced that subscriptions must be managed by a specific/separate service.

gunzip: what do you suggest in the meantime?

francescopersico: It is ok for me to merge as is.

francescopersico (approved):

PR #82, opened 2020-09-17
[#174850132] Prevent errors on missing notification

balanza:
This PR solves a bug for which, when a user request his/her data to be downloaded, if he/she has at leas one message with no related notification (this case is legal) the request fails.

### DIFF
--- a/ExtractUserDataActivity/handler.ts
+++ b/ExtractUserDataActivity/handler.ts
@@ -21,6 +21,7 @@ import {
   fromEither,
   TaskEither,
   taskEither,
+  taskEitherSeq,
   taskify,
   tryCatch
 } from "fp-ts/lib/TaskEither";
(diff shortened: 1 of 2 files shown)


ExtractUserDataActivity/handler.ts
```diff
@@ -292,32 +294,27 @@ export const findNotificationsForAllMessages = (
...
+            reason: `notificationModel.findNotificationForMessage| ${
+              e.kind
+            }, ${getMessageFromCosmosErrors(e)}`
+          })
+        )
       )
     )
-  );
+    .map(arrayOfMabeNotifications =>
+      arrayOfMabeNotifications
```
gunzip, line 311: you can use `array.catOptions` === filter(isSome) + x.value

https://github.com/gcanti/fp-ts/blob/1.17.4/src/Array.ts#L1312

ExtractUserDataActivity/handler.ts
```diff
@@ -292,32 +294,27 @@ export const findNotificationsForAllMessages = (
...
+            kind: "QUERY_FAILURE",
+            reason: `notificationModel.findNotificationForMessage| ${
+              e.kind
+            }, ${getMessageFromCosmosErrors(e)}`
+          })
+        )
       )
     )
-  );
+    .map(arrayOfMabeNotifications =>
```
gunzip, line 310:
```suggestion
    .map(array.catOptions) =>
```

PR #84, opened 2020-09-18
[#174823724] Add token name to GetUser from Azure ADB2C

AleDore:
This PR is useful to retrieve custom extension property `token_name` from Azure AD B2C, while calling `GetUser` API. Property value is also mapped into `UserInfo` API model in order to make possible info's propagation to io-functions-services.

### DIFF
--- a/GetUser/handler.ts
+++ b/GetUser/handler.ts
@@ -35,6 +35,7 @@ import { EmailAddress } from "../generated/definitions/EmailAddress";
 import { UserInfo } from "../generated/definitions/UserInfo";
 import {
   getApiClient,
+  getGraphRbacManagementClient,
   getUserGroups,
   IAzureApimConfig,
   IServicePrincipalCreds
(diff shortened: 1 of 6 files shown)


GetUser/handler.ts
```diff
@@ -193,6 +195,34 @@ export function GetUserHandler(
...
+                  error as Error
+                )
+            ).map(userList => userList[0])
+          )
+          .map(adb2User => {
+            return {
+              ...taskResults,
+              token_name:
+                adb2User[
+                  `adb2User.extension_${adb2cCredentials.clientId}_token_name`
```
gunzip, line 221: what if the user does not have this attribute ? (may you try to delete ti into the ADB2C and call GetUser?)

AleDore: I think that many users does not have this attribute on ADB2C. In that case token_name should be null and it will remain null until we add token_name on Active Directory. Why should we delete the user if it does not have `token_name`?

PR #85, opened 2020-09-18
[#17482372] Add token name to GetUser from Azure ADB2C

AleDore:
This PR is useful to retrieve custom extension property token_name from Azure AD B2C, while calling GetUser API. Property value is also mapped into UserInfo API model in order to make possible info's propagation to io-functions-services.

Edit: Fix on token attribute's naming


gunzip:
it  looks like there are some conflicts

gunzip (approved):

PR #89, opened 2020-09-30
[#175059122] Centralise config read

balanza:
This PR introduces the `utils/config` module which aims to be the unique interface for application configuration. 

Every place in which the application used to directly access env variable has been refactored to use this module instead.

### DIFF
--- a/DeleteUserDataActivity/index.ts
+++ b/DeleteUserDataActivity/index.ts
@@ -16,10 +16,9 @@ import { NotificationStatusDeletableModel } from "../utils/extensions/models/not
 import { ProfileDeletableModel } from "../utils/extensions/models/profile";
 
 const config = getConfigOrThrow();
-const cosmosDbName = config.COSMOSDB_NAME;
 
 const messagesContainer = cosmosdbClient
-  .database(cosmosDbName)
+  .database(config.COSMOSDB_NAME)
   .container(MESSAGE_COLLECTION_NAME);
 
 const messageModel = new MessageDeletableModel(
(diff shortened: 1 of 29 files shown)


gunzip:
lgtm, if you handle the FIXMEs we can proceed and merge it

AleDore (commented):
lgtm. Maybe we have to check if there are any case of "optional" env variable, and how we can handle it using this approach.

gunzip:
lgtm, why draft ?

DeleteUserDataActivity/index.ts
```diff
@@ -10,21 +8,23 @@ import { MESSAGE_STATUS_COLLECTION_NAME } from "io-functions-commons/dist/src/mo
...
 import { NotificationDeletableModel } from "../utils/extensions/models/notification";
 import { NotificationStatusDeletableModel } from "../utils/extensions/models/notification_status";
 import { ProfileDeletableModel } from "../utils/extensions/models/profile";
 
-const cosmosDbName = getRequiredStringEnv("COSMOSDB_NAME");
+const config = getConfigOrThrow();
+const cosmosDbName = config.COSMOSDB_NAME;
 
 const messagesContainer = cosmosdbClient
   .database(cosmosDbName)
```
BurnedMarshal, line 22: Can we use directly the config value?
```suggestion
const messagesContainer = cosmosdbClient
  .database(config.COSMOSDB_NAME)
```

BurnedMarshal (commented):
lgtm, only check if is better variables aliasing instead using directly the values

PR #90, opened 2020-09-30
[#175014199] Health checks

balanza:
Introducing health checks for the application.

* a `utils/healthcheck` module
* check for wrong configuration and unreachable resources (cosmosdb, storage, urls)
* add healthcheck on `/info` endpoint
* add `/info` endpoint to openapi spec

### DIFF
--- a/Info/__tests__/handler.test.ts
+++ b/Info/__tests__/handler.test.ts
@@ -1,29 +1,29 @@
+import { taskEither, fromLeft } from "fp-ts/lib/TaskEither";
+import { HealthCheck } from "../../utils/healthcheck";
 import { InfoHandler } from "../handler";
 
-const mockGetConfig = jest.fn(() => {});
+const mockHealthCheck = jest.fn<HealthCheck, []>(() => taskEither.of(true));
 
 afterEach(() => {
   jest.clearAllMocks();
 });
 
 describe("InfoHandler", () => {
-  it("should return an internal error if the configuration is wrong", async () => {
-    mockGetConfig.mockImplementationOnce(() => {
-      throw new Error("failure");
-    });
+  it("should return an internal error if the application is not healthy", async () => {
+    mockHealthCheck.mockImplementationOnce(() =>
+      fromLeft(["failure 1", "failure 2"])
+    );
 
-    // @ts-ignore
-    const handler = InfoHandler({ getConfig: mockGetConfig });
+    const handler = InfoHandler(mockHealthCheck);
 
     const response = await handler();
 
     expect(response.kind).toBe("IResponseErrorInternal");
   });
 
-  it("should return a success if the configuration is ok", async () => {
-    // @ts-ignore
-    const handler = InfoHandler({ getConfig: mockGetConfig });
-
+  it("should return a success if the application is healthy", async () => {
+    const handler = InfoHandler(mockHealthCheck);
+    
     const response = await handler();
 
     expect(response.kind).toBe("IResponseSuccessJson");
--- a/Info/__tests__/handler.test.ts
+++ b/Info/__tests__/handler.test.ts
@@ -2,19 +2,14 @@ import { taskEither, fromLeft } from "fp-ts/lib/TaskEither";
 import { HealthCheck } from "../../utils/healthcheck";
 import { InfoHandler } from "../handler";
 
-const mockHealthCheck = jest.fn<HealthCheck, []>(() => taskEither.of(true));
-
 afterEach(() => {
   jest.clearAllMocks();
 });
 
 describe("InfoHandler", () => {
   it("should return an internal error if the application is not healthy", async () => {
-    mockHealthCheck.mockImplementationOnce(() =>
-      fromLeft(["failure 1", "failure 2"])
-    );
-
-    const handler = InfoHandler(mockHealthCheck);
+    const healthCheck: HealthCheck = fromLeft(["failure 1", "failure 2"]);
+    const handler = InfoHandler(healthCheck);
 
     const response = await handler();
 
@@ -22,8 +17,9 @@ describe("InfoHandler", () => {
   });
 
   it("should return a success if the application is healthy", async () => {
-    const handler = InfoHandler(mockHealthCheck);
-    
+    const healthCheck: HealthCheck = taskEither.of(true);
+    const handler = InfoHandler(healthCheck);
+
     const response = await handler();
 
     expect(response.kind).toBe("IResponseSuccessJson");
--- a/Info/__tests__/handler.test.ts
+++ b/Info/__tests__/handler.test.ts
@@ -1,0 +1,31 @@
+import { InfoHandler } from "../handler";
+
+const mockGetConfig = jest.fn(() => {});
+
+afterEach(() => {
+  jest.clearAllMocks();
+});
+
+describe("InfoHandler", () => {
+  it("should return an internal error if the configuration is wrong", async () => {
+    mockGetConfig.mockImplementationOnce(() => {
+      throw new Error("failure");
+    });
+
+    // @ts-ignore
+    const handler = InfoHandler({ getConfig: mockGetConfig });
+
+    const response = await handler();
+
+    expect(response.kind).toBe("IResponseErrorInternal");
+  });
+
+  it("should return a success if the configuration is ok", async () => {
+    // @ts-ignore
+    const handler = InfoHandler({ getConfig: mockGetConfig });
+
+    const response = await handler();
+
+    expect(response.kind).toBe("IResponseSuccessJson");
+  });
+});
--- a/Info/__tests__/handler.test.ts
+++ b/Info/__tests__/handler.test.ts
@@ -1,5 +1,5 @@
 import { taskEither, fromLeft } from "fp-ts/lib/TaskEither";
-import { HealthCheck } from "../../utils/healthcheck";
+import { HealthCheck, HealthProblem } from "../../utils/healthcheck";
 import { InfoHandler } from "../handler";
 
 afterEach(() => {
@@ -8,7 +8,10 @@ afterEach(() => {
 
 describe("InfoHandler", () => {
   it("should return an internal error if the application is not healthy", async () => {
-    const healthCheck: HealthCheck = fromLeft(["failure 1", "failure 2"]);
+    const healthCheck: HealthCheck = fromLeft([
+      "failure 1" as HealthProblem<"Config">,
+      "failure 2" as HealthProblem<"Config">
+    ]);
     const handler = InfoHandler(healthCheck);
 
     const response = await handler();
--- a/Info/handler.ts
+++ b/Info/handler.ts
@@ -1,27 +1,47 @@
 import * as express from "express";
+import { toError, tryCatch2v } from "fp-ts/lib/Either";
 import { wrapRequestHandler } from "io-functions-commons/dist/src/utils/request_middleware";
 import {
   IResponseSuccessJson,
-  ResponseSuccessJson
+  ResponseSuccessJson,
+  IResponseErrorInternal,
+  ResponseErrorInternal
 } from "italia-ts-commons/lib/responses";
 import * as packageJson from "../package.json";
+import { IConfig } from "../utils/config";
 
 interface IInfo {
   version: string;
 }
 
-type InfoHandler = () => Promise<IResponseSuccessJson<IInfo>>;
+type InfoHandler = () => Promise<
+  IResponseSuccessJson<IInfo> | IResponseErrorInternal
+>;
 
-export function InfoHandler(): InfoHandler {
+export function InfoHandler({
+  getConfig
+}: {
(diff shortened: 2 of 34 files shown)


utils/config.ts
```diff
@@ -0,0 +1,97 @@
...
+ * The configuration is evaluate eagerly at the first access to the module. The module exposes convenient methods to access such value.
+ */
+
+import { tryCatch2v } from "fp-ts/lib/Either";
+import { toError } from "fp-ts/lib/Either";
+import { getRequiredStringEnv } from "io-functions-commons/dist/src/utils/env";
+
+// base read function
+// tslint:disable typedef
+const getEnv = () => ({
```
gunzip, line 14: an alternative to use `getRequiredStringEnv`  is to `decode(process.env)` using a type with (eventually) partials and (eventually) unions (in case of attributes dependencies). the actual code will fail on the first error, the decode can collect all errors (using Validation). aside, you don't have to type two times the name of the variables (which is error prone). what do you think about it?

Info/handler.ts
```diff
@@ -1,27 +1,37 @@
...
-      version: packageJson.version
-    });
-  };
+export function InfoHandler<T>(healthCheck: HealthCheck<T>): InfoHandler {
+  return () =>
+    healthCheck
+      .fold<IResponseSuccessJson<IInfo> | IResponseErrorInternal>(
+        problems => ResponseErrorInternal(problems.join("\n")),
+        _ =>
+          ResponseSuccessJson({
```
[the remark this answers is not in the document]
gunzip: something like this: https://tools.ietf.org/html/draft-inadarei-api-health-check-04
(but simpler)

utils/config.ts
balanza: good idea

utils/config.ts
```diff
@@ -0,0 +1,97 @@
+/**
```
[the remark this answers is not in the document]
balanza: Both this PR and #89 have just one file in common, but very different impact. I'd rather handle them separately and take myself the burden of keeping the lone file in sync.

Info/handler.ts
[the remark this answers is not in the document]
balanza: > return all the statuses even in case of success

What d'you mean? We are checking only the current application status - at most, if it can reach an external resources. So I think we just have a single pass/fail status

> common vocabulary for status checks.

I agree indeed.

> something like this: https://tools.ietf.org/html/draft-inadarei-api-health-check-04

I like this approach, but I notice that the above is a draft. Meanwhile, we are using a `problem+json` which is already a standard: https://tools.ietf.org/html/rfc7807. 

As for now, we need:
* a machine-readable encoding of the service pass/fail status
   |> we use http status code for that
* a human-readable report of failures, for throubleshooting
   |> we user `detail` section of the `ProblemJson` schema for that

We can model further and define a proper encoding for failure reports, to make them machine-readable. However, I see no use for that so far.

utils/healthcheck.ts
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Check a url is reachable
+ *
+ * @param url url to connect with
+ *
+ * @returns either true or an array of error messages
+ */
+export const checkUrlHealth = (_: string): HealthCheck =>
+  // TODO: implement this check
+  taskEither.of(true);
```
balanza, line 76: These functions may be moved into `io-functions-commons` (or even io-ts-commons)

gunzip: functions-commons is ok (ts-commons is used by the app as well)

utils/config.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    PUBLIC_DOWNLOAD_BASE_URL: NonEmptyString,
+
+    SESSION_API_KEY: NonEmptyString,
+    SESSION_API_URL: NonEmptyString,
+
+    LOGOS_URL: NonEmptyString,
+
+    SUBSCRIPTIONS_FEED_TABLE: NonEmptyString,
+    USER_DATA_DELETE_DELAY_DAYS: NonEmptyString,
```
AleDore, line 132: We have to remember that in case of a global env variable creation or deletion or renaming we have to update this

AleDore (commented):
lgtm, apart form minor discussion

utils/__tests__/config.test.ts
```diff
@@ -0,0 +1,199 @@
...
+      NODE_ENV: "production",
+      SENDGRID_API_KEY: "a-sg-key"
+    };
+    const result = MailerConfig.decode(rawConf);
+
+    expectRight(result, value => {
+      expect(value.SENDGRID_API_KEY).toBe("a-sg-key");
+      expect(typeof value.MAILUP_USERNAME).toBe("undefined");
+    });
+  });
```
BurnedMarshal, line 49: This is a duplicate test

utils/__tests__/config.test.ts
```diff
@@ -0,0 +1,199 @@
...
+      MAILUP_SECRET: "a-mu-secret"
+    };
+    const result = MailerConfig.decode(rawConf);
+
+    expectRight(result, value => {
+      expect(value.MAILUP_USERNAME).toBe("a-mu-username");
+      expect(value.MAILUP_SECRET).toBe("a-mu-secret");
+      // check types
+      const _: NonEmptyString = value.MAILUP_SECRET;
+      const __: NonEmptyString = value.MAILUP_USERNAME;
```
BurnedMarshal, line 80: This can be removed

Info/handler.ts
```diff
@@ -1,27 +1,37 @@
...
-  return async () => {
-    return ResponseSuccessJson({
-      version: packageJson.version
-    });
-  };
+export function InfoHandler(healthCheck: HealthCheck): InfoHandler {
+  return () =>
+    healthCheck
+      .fold<IResponseSuccessJson<IInfo> | IResponseErrorInternal>(
+        problems => ResponseErrorInternal(problems.join("\n\n")),
```
BurnedMarshal, line 24: Are we sure than any critical information come from `healthCheck` is filtered before sending it through the public API `Info`?

utils/__tests__/config.test.ts
```diff
@@ -0,0 +1,199 @@
...
+      fail(`Expecting right, received left. Value: ${JSON.stringify(e.value)}`),
+    _ => t(_)
+  );
+
+const expectLeft = <L, R>(e: Either<L, R>, t: (l: L) => void = noop) =>
+  e.fold(
+    _ => t(_),
+    _ =>
+      fail(`Expecting left, received right. Value: ${JSON.stringify(e.value)}`)
+  );
```
BurnedMarshal, line 20: Nice approach. Can we move these methods to a test utility library to share this logic across all the projects?

Info/__tests__/handler.test.ts
```diff
@@ -0,0 +1,27 @@
...
+import { HealthCheck } from "../../utils/healthcheck";
+import { InfoHandler } from "../handler";
+
+afterEach(() => {
+  jest.clearAllMocks();
+});
+
+describe("InfoHandler", () => {
+  it("should return an internal error if the application is not healthy", async () => {
+    const healthCheck: HealthCheck = fromLeft(["failure 1", "failure 2"]);
```
BurnedMarshal, line 11: There is a type error in here, the test fails.

utils/healthcheck.ts
```diff
@@ -0,0 +1,130 @@
...
+  }, toHealthProblems("AzureCosmosDB")).map(_ => true);
+
+/**
+ * Check the application can connect to an Azure Storage
+ *
+ * @param connStr connection string for the storage
+ *
+ * @returns either true or an array of error messages
+ */
+export const checkAzureStorageHealth = (
```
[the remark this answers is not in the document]
balanza: Good point. This checks the storage resource itself and if the app can correctly connect with it. However I actually found no better solution than try to connect to a blob storage.

utils/__tests__/config.test.ts
balanza: Do yo mean the whole case or just `expect(typeof value.MAILUP_USERNAME).toBe("undefined");`?

utils/healthcheck.ts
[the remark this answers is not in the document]
balanza: I'd leave the scope of the test to be "the whole storage resource" (thus, I'd leave the name) and I'd rather improve the implementation. Suggestions?

utils/__tests__/config.test.ts
balanza: This is actually a canary to catch possible problems when/if we'll edit the config model. I found that several times the programmatic check doesn't ensure the correct type inference (hence, you have green tests but the project will fail to build). Not exhaustive, though.

Info/handler.ts
balanza: Good point. For the case I tested I saw no problems (resource unavailable, wrong credentials). Still it's important to point out, thanks. 

Do you see any case which can be problematic? Please also consider that, for what concerns to functions, the endpoint is unauthenticated but not public, it's reachable only from our private network.

utils/__tests__/config.test.ts
BurnedMarshal: The whole test is exactly the same as the previous one

Info/handler.ts
BurnedMarshal: We could anyway return a generic Error message of what type of error occurs `problems.map(_ => _.__source);` and logs the complete error for detail. What do you think?

utils/__tests__/config.test.ts
balanza: I'd like to write a jest extension module with custom matchers for io-ts/fp-ts. Maybe one day ;)
There are still quirks with the above functions (failures aren't reported very well)

utils/healthcheck.ts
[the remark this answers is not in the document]
BurnedMarshal: We could extend the check on lower resources es. for Blob check that the containers exist ([api](https://docs.microsoft.com/en-us/rest/api/storageservices/get-container-properties)), for table storage that all required tables exist and so on. `checkAzureStorageHealth` should regroup all these storage health checks. Attention on autogenerated resources.

utils/__tests__/config.test.ts
BurnedMarshal: Sorry, I don't understand. At runtime we have ```IConfig.decode({
  ...process.env,
  isProduction: process.env.NODE_ENV === "production"
});``` to check the correct type of required values.

Info/handler.ts
balanza: lgtm as soon as it'll be easy to troubleshoot. @gunzip thoughts?

utils/__tests__/config.test.ts
balanza: oops, didn't notice that. Good catch!

utils/__tests__/config.test.ts
BurnedMarshal: If you need a type check inside the tests is better to create an explicit type
```
const MailUpConfig = t.interface({
    MAILHOG_HOSTNAME: t.undefined,
    MAILUP_SECRET: NonEmptyString,
    MAILUP_USERNAME: NonEmptyString,
    MAIL_TRANSPORTS: t.undefined,
    NODE_ENV: t.literal("production"),
    SENDGRID_API_KEY: t.undefined
})
```
And check the response. We can use these types even inside the logic with `MailUpConfig.is(config)`

utils/config.ts
balanza: No problem in creation - you'll be forced to add it to the model. When deleting it is tricky, and we'd must rely on our discipline imho

utils/healthcheck.ts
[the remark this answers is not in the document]
balanza: >check that the containers exist 

That is a deeper level of check which will introduce additional considerations, for example many containers are lazily created. 

>Attention on autogenerated resources.

What do you mean?

utils/__tests__/config.test.ts
balanza: Well, the check is all about "let me se if the interface I designed actually results in the correct type once I exit the io-ts world".
In the case above, it happened that `MAILUP_USERNAME` could not be correctly narrowed to `NonEmptyString` but to `string | NonEmptyString`, although the interface was programmatically good.
Maybe I could do `NonEmptyString.is(value.MAILUP_USERNAME)`

utils/healthcheck.ts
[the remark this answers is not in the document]
BurnedMarshal: > What do you mean?
The same as you for "lazily created containers". They don't need to be checked. Must be checked only required resource for application start and run. We can improve this logic later if hard to implements.

gunzip:
once rebased, is this ready to be merged?

balanza:
> once rebased, is this ready to be merged?

yes

Info/handler.ts
gunzip: 👍

gunzip (approved):

PR #91, opened 2020-10-07
[#175170499] Add missing unauthorized to openapi specs

AleDore:
This PR adds missing unauthorized response to operations that could return 401, with focus on `GetUser` and `CreateSubscription`


gunzip (approved):
thank you @AleDore ! kudos to @balanza as the generated clients are helping us fixing broken specs :-D

PR #92, opened 2020-10-12
[#175227073] Allow updating ADB2C User and introduce Token Name management on Create/Update operations

AleDore:
This PR is intended to manage ADB2C custom attribute `Token Name`, in order to allow existing `CreateUser` and the brand new `UpdateUser` to manage this custom attribute.

### DIFF
--- a/UpdateUser/__tests__/handler.ts
+++ b/UpdateUser/__tests__/handler.ts
@@ -3,9 +3,11 @@
 import { GraphRbacManagementClient } from "@azure/graph";
 import * as msRestNodeAuth from "@azure/ms-rest-nodeauth";
 import { NonEmptyString } from "italia-ts-commons/lib/strings";
+import { EmailAddress } from "../../generated/definitions/EmailAddress";
 import { UserCreated } from "../../generated/definitions/UserCreated";
 import { UserPayload } from "../../generated/definitions/UserPayload";
 import { UserStateEnum } from "../../generated/definitions/UserState";
+import { UserUpdatePayload } from "../../generated/definitions/UserUpdatePayload";
 import { IServicePrincipalCreds } from "../../utils/apim";
 import { UpdateUserHandler } from "../handler";
 
@@ -16,12 +18,12 @@ const fakeServicePrincipalCredentials: IServicePrincipalCreds = {
   tenantId: "tenant-id"
 };
 
+const aUserEmail = "user@example.com" as EmailAddress;
 const fakeRequestPayload = {
-  email: "user@example.com",
   first_name: "first-name",
   last_name: "family-name",
   token_name: aTokenName
-} as UserPayload;
+} as UserUpdatePayload;
 
 const fakeObjectId = "ADB2C-user";
 
@@ -75,6 +77,7 @@ describe("UpdateUser", () => {
     const response = await updateUserHandler(
       mockedContext as any,
       undefined as any,
+      undefined as any,
       undefined as any
     );
 
@@ -94,6 +97,7 @@ describe("UpdateUser", () => {
     const response = await updateUserHandler(
       mockedContext as any,
       undefined as any,
+      aUserEmail,
       fakeRequestPayload
     );
 
@@ -102,7 +106,7 @@ describe("UpdateUser", () => {
 
   it("should return the user updated", async () => {
     const fakeApimUser = {
-      email: fakeRequestPayload.email,
+      email: aUserEmail,
       firstName: fakeRequestPayload.first_name,
       id: "user-id",
       identities: [
@@ -136,6 +140,7 @@ describe("UpdateUser", () => {
     const response = await updateUserHandler(
       mockedContext as any,
       undefined as any,
+      aUserEmail,
       fakeRequestPayload
     );
     expect(response).toEqual({
--- a/UpdateUser/__tests__/handler.ts
+++ b/UpdateUser/__tests__/handler.ts
@@ -5,7 +5,6 @@ import * as msRestNodeAuth from "@azure/ms-rest-nodeauth";
 import { NonEmptyString } from "italia-ts-commons/lib/strings";
 import { EmailAddress } from "../../generated/definitions/EmailAddress";
 import { UserCreated } from "../../generated/definitions/UserCreated";
-import { UserPayload } from "../../generated/definitions/UserPayload";
 import { UserStateEnum } from "../../generated/definitions/UserState";
 import { UserUpdatePayload } from "../../generated/definitions/UserUpdatePayload";
 import { IServicePrincipalCreds } from "../../utils/apim";
--- a/UpdateUser/__tests__/handler.ts
+++ b/UpdateUser/__tests__/handler.ts
@@ -1,0 +1,147 @@
+// tslint:disable:no-any
+
+import { GraphRbacManagementClient } from "@azure/graph";
+import * as msRestNodeAuth from "@azure/ms-rest-nodeauth";
+import { NonEmptyString } from "italia-ts-commons/lib/strings";
+import { UserCreated } from "../../generated/definitions/UserCreated";
+import { UserPayload } from "../../generated/definitions/UserPayload";
+import { UserStateEnum } from "../../generated/definitions/UserState";
+import { IServicePrincipalCreds } from "../../utils/apim";
+import { UpdateUserHandler } from "../handler";
+
+const aTokenName = "ATokenName" as NonEmptyString;
+const fakeServicePrincipalCredentials: IServicePrincipalCreds = {
+  clientId: "client-id",
+  secret: "secret",
+  tenantId: "tenant-id"
+};
+
+const fakeRequestPayload = {
+  email: "user@example.com",
+  first_name: "first-name",
+  last_name: "family-name",
+  token_name: aTokenName
+} as UserPayload;
+
+const fakeObjectId = "ADB2C-user";
+
+const mockLoginWithServicePrincipalSecret = jest.spyOn(
+  msRestNodeAuth,
+  "loginWithServicePrincipalSecret"
+);
+
+jest.mock("@azure/graph");
+jest.mock("@azure/arm-apimanagement");
+const mockGraphRbacManagementClient = GraphRbacManagementClient as jest.Mock;
+const mockLog = jest.fn();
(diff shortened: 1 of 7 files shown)


francescopersico:
When you update a user, setting a different token_name, what happens to already created service? The token_name in the services remains the old one. Can it be a problem?

gunzip:
The ratio here is that attributes on user account are _default_ values for _newly_ create services, so I think it's not a problem if the old services maintain the old token name (it must be documented)

AleDore:
> When you update a user, setting a different token_name, what happens to already created service? The token_name in the services remains the old one. Can it be a problem?

Since this operation will be performed only by administrators, I think it will be used only for newly accounts or on existing ones, that will be enabled after the test's phase. It shouldn't be a problem if old services maintain the old token name.

UpdateUser/handler.ts
```diff
@@ -0,0 +1,133 @@
...
+        "UpdateUser | " + errorMessage,
+        error,
+        errorMessage
+      );
+    return getGraphRbacManagementClient(adb2cCredentials)
+      .mapLeft(error =>
+        internalErrorHandler("Could not get the ADB2C client", error)
+      )
+      .chain(graphRbacManagementClient =>
+        getUserFromList(graphRbacManagementClient, userPayload.email)
```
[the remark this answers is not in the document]
AleDore: Yep, `userPayload` is validated [here](https://github.com/pagopa/io-functions-admin/blob/f5dc43abb6352ef22c715bac8305e5e8e42c644e/UpdateUser/handler.ts#L129)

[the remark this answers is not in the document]
balanza: sorry didn't see that

AleDore (commented):
Can we merge this? @francescopersico @gunzip ?

AleDore:
I' ve checked that on `updateUser` operation, we expect `first_name` and `last_name` in the `userPayload`, but I think that the only information we have about the user to update is `email` isn't it? cc: @gunzip @francescopersico

AleDore:
> I think updateUser should work in the same way as the other update methods (ie. UpdateService), retrieving the old user properties and merging the overridden values, what about it?

Nope because graph API expect a payload containing only the attributes to be updated (so other attributes are not overwritten if not present ), but the question is: Do we always have informations about `first_name` and `last_name` while performing `UpdateUser` operation ? I think that the only information we got in that case is the `email` associated with the B2C account, so first_name and last_name should be optional in the userPayload used for Update. What do you think?

gunzip:
that's right, maybe you can use` t.partial(User)`for the decoder

AleDore:
> that's right, maybe you can use` t.partial(User)`for the decoder

Maybe we can define a payload for `UpdateUser` that has all possible attributes optional.

UpdateUser/handler.ts
```diff
@@ -0,0 +1,148 @@
...
+    toError
+  ).map(
+    updateUserResponse =>
+      ({
+        email,
+        first_name: userPayload.first_name
+          ? userPayload.first_name
+          : user.givenName,
+        id: updateUserResponse.objectId,
+        last_name: userPayload.last_name ? userPayload.last_name : user.surname,
```
gunzip, line 80: why this mapping here? can't we just return the exact values ?
first_name -> first_name, last_name -> last_name

AleDore: `UserCreated` has `first_name` and `last_name` as required fields. Since `userPayload` has the same optional attributes, we can map over `user` attributes as a fallback option.

balanza: we can make it shorter like 

```ts
first_name: userPayload.first_name || user.givenName
```

less cognitive load imho

gunzip: why do we need the fallback?

UpdateUser/handler.ts
```diff
@@ -0,0 +1,148 @@
...
+        token_name: userPayload.token_name
+      } as UserCreated)
```
balanza, line 82: Instead of this manual cast, can should we use a codec?

UpdateUser/handler.ts
AleDore: > why do we need the fallback? moreover, it prevent the caller to check if the values are undefined

Because if the caller doesn't provide `first_name` or `last_name` (namely it doesn't want to update first_name or last_name), we must provide the same attributes on `UserCreated` response because they are required. I think that B2C user always has this attributes ('cause createUser expects them in the request payload as required fields)

UpdateUser/handler.ts
```diff
@@ -69,17 +70,14 @@ const updateUser = (
...
-        token_name: userPayload.token_name
-      } as UserCreated)
+  ).chain(updateUserResponse =>
+    UserCreated.decode({
+      email,
+      first_name: userPayload.first_name || user.givenName,
+      id: updateUserResponse.objectId,
+      last_name: userPayload.last_name || user.surname,
+      token_name: userPayload.token_name
+    }).fold(errs => fromLeft(toError(errs)), usr => taskEither.of(usr))
```
balanza, line 80: what about 

```suggestion
  ).chain(updateUserResponse =>
    fromEither(UserCreated.decode({
      email,
      first_name: userPayload.first_name || user.givenName,
      id: updateUserResponse.objectId,
      last_name: userPayload.last_name || user.surname,
      token_name: userPayload.token_name
    }))
```

UpdateUser/handler.ts
```diff
@@ -4,7 +4,8 @@ import { User } from "@azure/graph/esm/models";
 import * as express from "express";
 import { toError } from "fp-ts/lib/Either";
 import { identity } from "fp-ts/lib/function";
-import { TaskEither, tryCatch } from "fp-ts/lib/TaskEither";
+import { fromLeft } from "fp-ts/lib/TaskEither";
+import { taskEither, TaskEither, tryCatch } from "fp-ts/lib/TaskEither";
```
balanza, line 8: If you accept [that](https://github.com/pagopa/io-functions-admin/pull/92/files/fcaa6c49c6d6c31c93e5ab9bed0fbdef5e809c04..924224cb476947ebace8390e87d0425b7b69c5ad#r504752786) suggestion you'll need this, too

```suggestion
import { fromEither, taskEither, TaskEither, tryCatch } from "fp-ts/lib/TaskEither";
```

UpdateUser/handler.ts
gunzip: fallback is not a good idea (it's an arbitrary mapping just to fill value to satisfy the decoder). if we must return these values then they must be retrieved from the original user record otherwise we should relax the constraints on these fields

UpdateUser/handler.ts
balanza, line 8: If you accept [that](https://github.com/pagopa/io-functions-admin/pull/92/files/fcaa6c49c6d6c31c93e5ab9bed0fbdef5e809c04..924224cb476947ebace8390e87d0425b7b69c5ad#r504752786) suggestion you'll need this, too

```suggestion
import { fromEither, taskEither, TaskEither, tryCatch } from "fp-ts/lib/TaskEither";
```

AleDore: My bad, you' re right 😄

UpdateUser/handler.ts
AleDore: Ok let's create a `UserUpdated` instead of `UserCreated` that allows direct mapping from `userPayload`

UpdateUser/handler.ts
```diff
@@ -4,8 +4,8 @@ import { User } from "@azure/graph/esm/models";
 import * as express from "express";
 import { toError } from "fp-ts/lib/Either";
 import { identity } from "fp-ts/lib/function";
-import { fromLeft } from "fp-ts/lib/TaskEither";
-import { taskEither, TaskEither, tryCatch } from "fp-ts/lib/TaskEither";
+import { fromEither } from "fp-ts/lib/TaskEither";
+import { TaskEither, tryCatch } from "fp-ts/lib/TaskEither";
```
balanza, line 8: no need to import twice
```suggestion
import { fromEither, TaskEither, tryCatch } from "fp-ts/lib/TaskEither";
```

AleDore: It' s so strange that lint doesn't report a non grouped import 😄

AleDore:
Can we merge this? @balanza @BurnedMarshal

balanza (approved):

UpdateUser/__tests__/handler.ts
```diff
@@ -0,0 +1,151 @@
...
+const mockLog = jest.fn();
+const mockGetToken = jest.fn();
+
+mockLoginWithServicePrincipalSecret.mockImplementation(() => {
+  return Promise.resolve({ getToken: mockGetToken });
+});
+mockGetToken.mockImplementation(() => {
+  return Promise.resolve(undefined);
+});
+const mockUsersCreate = jest.fn();
```
BurnedMarshal, line 46: Could be this `mockUsersUpdate`?

UpdateUser/__tests__/handler.ts
```diff
@@ -0,0 +1,151 @@
...
+    );
+
+    const response = await updateUserHandler(
+      mockedContext as any,
+      undefined as any,
+      aUserEmail,
+      fakeRequestPayload
+    );
+
+    expect(response.kind).toEqual("IResponseErrorInternal");
```
BurnedMarshal, line 103: Can we relate the behaviour of this test with the test conditions?
For example 
```suggestion
    expect(mockUsersCreate).toBeCalledTimes(1);
    expect(response).toEqual({
      apply: expect.any(Function),
      detail: expectedError,
      kind: "IResponseErrorInternal"
    });
```

AleDore: Fixed in [2948303](https://github.com/pagopa/io-functions-admin/pull/92/commits/29483031cd2370eecdac4641c2bd3c7c1844805c)

UpdateUser/__tests__/handler.ts
AleDore: Fixed in [2948303](https://github.com/pagopa/io-functions-admin/pull/92/commits/29483031cd2370eecdac4641c2bd3c7c1844805c)

BurnedMarshal (approved):
lgtm

PR #95, opened 2020-10-21
[#175374436] Allow lowercase transformation on png naming while uploading a Service image Logo

AleDore:
This PR introduce lowercase transformation on blob name while uploading a service logo, due to wrong discovery by IO-APP that uses lowercase naming convention.
It also introduce the azure storage' sdk usage, because these kind of transformations (lowercase of the variables) are not allowed in azure bindings approach.

So changes can be resumed as follow:
* Out binding deletion (blob type) on `function.json`;
* Handler now use `blobService`, in order to write blob on `AssetsStorage` identified by `AssetsStorageConnection` config variable;
* Config module now handle `AssetsStorageConnection` that was missing;
* Tests refactoring due to implementation changes;

### DIFF
--- a/UploadServiceLogo/handler.ts
+++ b/UploadServiceLogo/handler.ts
@@ -106,6 +106,7 @@ export function UpdateServiceLogoHandler(
     }
 
     const bufferImage = Buffer.from(logoPayload.logo, "base64");
+    const lowerCaseServiceId = serviceId.toLowerCase();
     return fromEither(
       tryCatch(() => UPNG.decode(bufferImage)).foldL(
         () =>
@@ -130,7 +131,7 @@ export function UpdateServiceLogoHandler(
           tUpsertBlobFromObject(
             blobService,
             "services",
-            `${serviceId.toLowerCase()}.png`,
+            `${lowerCaseServiceId}.png`,
             bufferImage.toString()
           )
             .mapLeft(err =>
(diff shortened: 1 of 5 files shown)


UploadServiceLogo/index.ts
```diff
@@ -27,6 +28,8 @@ const servicesContainer = database.container(SERVICE_COLLECTION_NAME);
 
 const serviceModel = new ServiceModel(servicesContainer);
 
+const blobService = createBlobService(config.StorageConnection);
```
gunzip, line 31: isnt this `AssetsStorageConnection`?

AleDore: Yes it is but it's not present in config module :) I'd rather add this variable.

UploadServiceLogo/handler.ts
```diff
@@ -82,27 +106,63 @@ export function UpdateServiceLogoHandler(
...
+      .foldTaskEither<
+        IResponseErrorValidation | IResponseErrorInternal,
+        IResponseSuccessRedirectToResource<{}, {}>
+      >(
+        imageValidationError => fromLeft(imageValidationError),
+        () =>
+          tUpsertBlobFromObject(
+            blobService,
+            "services",
+            `${serviceId.toLowerCase()}.png`,
```
balanza, line 133: I'd save this value in a variable as it's being used later in the code.

balanza (commented):
It's not clear to me what changes have been introduced in this PR. The title and the related story both mention to keep the name lowercase, but then the while upload process has been revised.

AleDore:
> It's not clear to me what changes have been introduced in this PR. The title and the related story both mention to keep the name lowercase, but then the while upload process has been revised.

I've updated description :)

balanza:
Lgtm

balanza (approved):

AleDore:
Ok let's wait before merging this, due to integration tests execution.

PR #96, opened 2020-10-26
[#175395536] Add upload Organization Logo

AleDore:
This PR adds a brand new operation: `UploadOrganizationLogo`, that is similar to `ÙploadServiceLogo` but it is intended only for organization logos, that can be different from services logos.


balanza:
I'm thinking about api design and keep our interface in good shape.
 
I see we introduced the `/organizations` sub path for our endpoints. This let me think we now have a new concept, `Organization`, and it might make sense to have concept-specific functionalities such as _getOrganization_, _createOrganization_, etc (*). 
Is it the case? if so, I'm fine with the decision of organising endpoints like that. Otherwise, wouldn't make more sense to add `organization` as a sub-case for logo uploading (using a flag or a sub path)?



(*) I'm not saying we have to, I'm just exploring the domain.

AleDore:
> I'm thinking about api design and keep our interface in good shape.
> 
> I see we introduced the `/organizations` sub path for our endpoints. This let me think we now have a new concept, `Organization`, and it might make sense to have concept-specific functionalities such as _getOrganization_, _createOrganization_, etc (*).
> Is it the case? if so, I'm fine with the decision of organising endpoints like that. Otherwise, wouldn't make more sense to add `organization` as a sub-case for logo uploading (using a flag or a sub path)?
> 
> (*) I'm not saying we have to, I'm just exploring the domain.

It's a good point. I've used `/organizations` because potentially it's completely different from a service. An organization can create 1 or more service/s, but I agree with your point of view cc: @gunzip @francescopersico

gunzip:
there's some long term plan to have organizations as modelled entities so it's ok to have such an endpoint

balanza (approved):
LGTM

PR #100, opened 2020-10-29
[#175496403] Removed authentication and standard endpoint for health check

pasqualedevita:


Info/function.json
```diff
@@ -1,11 +1,11 @@
 {
   "bindings": [
     {
-      "authLevel": "function",
+      "authLevel": "anonymous",
       "type": "httpTrigger",
       "direction": "in",
       "name": "req",
-      "route": "adm/info",
+      "route": "info",
```
balanza, line 8: do we need this or is just a matter of style?

pasqualedevita: I think that is a good practice have the same standard API for all our health checks.
We need to expose info API on api management? I don't see any other info API configured on api management.

gunzip: If we don't have the need to avoid a conflict on the `/info` path (read: if we don't want to expose the API though an API managament) then you can strip down the prefix

gunzip (approved):

PR #105, opened 2020-11-05
[#175594485] Fix healthcheck gate deploy pipeline

pasqualedevita:
- add new parameter **containerInstanceResourceGroup**
- add **checks** if exists target vnet and subnet to avoid to create new vnet or subnet

### DIFF
--- a/azure-templates/deploy-steps.yml
+++ b/azure-templates/deploy-steps.yml
@@ -149,7 +149,7 @@ steps:
             exit 1
           fi
 
-          # check if subet exists in target resource group and vnet (containerInstanceResourceGroup, containerInstanceVNet)
+          # check if subnet exists in target resource group and vnet (containerInstanceResourceGroup, containerInstanceVNet)
           OUTPUT_SUBNET=$(az network vnet subnet list --resource-group '${{ parameters.containerInstanceResourceGroup }}' --vnet-name '${{ parameters.containerInstanceVNet }}' --query "[?contains(name, '${{ parameters.containerInstanceSubnet }}')].name" --output tsv)
           if [ "$OUTPUT_SUBNET" = '${{ parameters.containerInstanceSubnet }}' ]; then
             echo "SUBNET '${{ parameters.containerInstanceSubnet }}' found, continue"
--- a/azure-templates/deploy-steps.yml
+++ b/azure-templates/deploy-steps.yml
@@ -37,6 +37,11 @@ parameters:
     type: string
     default: '0.5'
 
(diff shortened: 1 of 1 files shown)


azure-templates/deploy-steps.yml
```diff
@@ -135,9 +140,27 @@ steps:
...
+          # check if vnet exists in target resource group (containerInstanceResourceGroup)
+          OUTPUT_VNET=$(az network vnet list --resource-group '${{ parameters.containerInstanceResourceGroup }}' --query "[?contains(name, '${{ parameters.containerInstanceVNet }}')].name" --output tsv)
+          if [ "$OUTPUT_VNET" = '${{ parameters.containerInstanceVNet }}' ]; then
+            echo "VNET '${{ parameters.containerInstanceVNet }}' found, continue"
+          else
+            echo "VNET '${{ parameters.containerInstanceVNet }}' not found, fail"
+            exit 1
+          fi
+
+          # check if subet exists in target resource group and vnet (containerInstanceResourceGroup, containerInstanceVNet)
```
gunzip, line 152:
```suggestion
          # check if subnet exists in target resource group and vnet (containerInstanceResourceGroup, containerInstanceVNet)
```

gunzip (approved):
lgtm cc @balanza

azure-templates/deploy-steps.yml
```diff
@@ -37,6 +37,11 @@ parameters:
     type: string
     default: '0.5'
 
+  # the resource group must be the same where vnet is created
+  - name: 'containerInstanceResourceGroup'
+    type: string
+    default: 'io-p-rg-common'
```
balanza, line 43: Can't we get this value from the environment?

balanza (changes requested):

azure-templates/deploy-steps.yml
pasqualedevita: thanks

azure-templates/deploy-steps.yml
pasqualedevita: `az network vnet subnet list` requires resource group to list subnets, so we can't get values from other resources.

I don't think that we'll changes these values often.

```
  # the resource group must be the same where vnet is created
  - name: 'containerInstanceResourceGroup'
    type: string
    default: 'io-p-rg-common'
  # attached vnet to the container instance
  - name: 'containerInstanceVNet'
    type: string
    default: 'io-p-vnet-common'
  # container instance subnet
  - name: 'containerInstanceSubnet'
    type: string
    default: 'azure-devops'
```

PR #107, opened 2020-11-06
[#175476527] Add release stage to pipeline

balanza:
Introduce a release stage to the deploy pipeline that does the following:
1. bump the version
2. fill the changelog
3. commit and tag 
4. push to the remote 
5. create a release on Github

## Key concepts
* Each release is tagged with `v{version}-RELEASE` and the bundle uploaded on Github releases.
* We assume the relation _one release <--> one deploy_.
* New developments can be only deployed from `master`. In such case, a new release is automatically bundled.
* Old versions of the app can be deployed by referencing a `v{version}-RELEASE` tag. Release won't be bundled as we assume it has been done already.

## Variables introduced
|name|description|default|
|-|-|-|
|DO_RELEASE|whether the release stage is enabled|true|
|GIT_EMAIL|Email of the user which commits to the repo|****|
|GIT_USERNAME|Name of the user which commits to the repo|`pagopa-github-bot`|
|GITHUB_CONNECTION|The name of the service connection used to authenticate on Github when pushing|****|
|RELEASE_SEMVER|The kind of release, either `major`, `minor` or `patch`. To be provided when the pipeline is started|`minor`|

#### Scenario 1: deploy the current master
1. select `Run pipeline`
1. leave `master` on `Branch/tag` field (should be default)
1. select the proper semver from the radio button list
1. click `Run`

#### Scenario 2: deploy a release that already exists
1. select `Run pipeline`
1. set `refs/tags/my-version-tag` on `Branch/tag` field 
1. ignore the semver param
1. click `Run` 

#### Scenario 3: disable release stage as default
1. select `Edit`
1. select `Variables`
1. set `DO_RELEASE` to false

## Changes to other stages
* A manual checkout is needed in the build steps in order to be sure to point to the very latest commit. This is needed because a stage checks out the repository at the commit on which the pipeline is started. When such commit is not the latest anymore (as it is when commit the version bump), you need to re-checkout.

## About versioning and changelog
* Package `release-it` is no longer needed as we do version bump using `npm version` and thus it has been removed along with `.release-it.js` file.
* Package `auto-changelog` is now needed as first level dependency (it used to be a dependency of `release-it`)
* As we tag twice, we have to instruct `auto-changelog` which tag to consider when parsing the history ([here](https://github.com/pagopa/io-functions-admin/pull/107/commits/bd8a40856cb1ca99206bcc0c906cdd863ebaa6fc#diff-84e92790952d9ab73f43145eeaf4729d7ff911e102f40d64397a8c71d33cbae6R5)). 
* `CHANGELOG.md` is updated by the `version` script in the `package.json` file. A `preversion` script is kept for visualising item that will be added.
* The `Release` stage isn't aware changelog's stuff: it's the application's duty to update it.


azure-pipelines.yml
```diff
@@ -168,8 +181,39 @@ stages:
...
+      vmImage: 'ubuntu-latest'
+    jobs:
+      - job: make_release      
+        steps:
+        - template: azure-templates/make-release-steps.yml
+          parameters:
+            semver: '${{ parameters.RELEASE_SEMVER }}'
+            gitEmail: $(GIT_EMAIL)
+            gitUsername: $(GIT_USERNAME)
+            gitHubConnection: $(GITHUB_CONNECTION)
```
gunzip, line 214: can we set these ones at organization level ? to avoid setting these for every pipeline / project

azure-pipelines.yml
```diff
@@ -224,6 +270,7 @@ stages:
       dependsOn:
         - Build
```
gunzip, line 271: do we still need the build step here? I think that the deploy step makes a build itself (so probably we should remove this one)

azure-templates/make-release-steps.yml
```diff
@@ -0,0 +1,81 @@
...
+
+- script: |
+    git push --follow-tags origin master 
+  displayName: 'Push to master'
+
+- script: |
+    NEXT_VERSION=$(node -p "require('./package.json').version")
+    HEAD_SHA=$(git rev-parse HEAD)
+    TITLE="Release $NEXT_VERSION"
+    TAG="v$NEXT_VERSION"
```
gunzip, line 66: can we add some suffix to the release tags (in order to detect them among all the other possible tags)
ie. `-RELEASE` ?

azure-pipelines.yml
balanza: The build step is meant ton check if changes of a PR break the build. What we can do is to run them with `--noEmit` to save some processing, maybe.

Anyway, by introducing the Release stage, the Build and Deploy stages actually compile different source codes, so we must keep them separated (unless we do some dirty patching on built code).

azure-templates/make-release-steps.yml
balanza: Here is not where we create the tag, this is just how we instruct the Github release task on what tag point to. At this step, tag is already created and pushed. 

Tag is actually created automatically by `npm version`. If we need more control over the format, we have a couple of options:
1. use [`git-tag-version`](https://docs.npmjs.com/cli/v6/commands/npm-version#git-tag-version) flag on the command, which blocks commit and tagging (which we have to do manually then)
1. use [tag-version-prefix](https://docs.npmjs.com/cli/v6/using-npm/config#tag-version-prefix) config, but it actually works on prefix only.

I'd go for the first, but **do we really need to do that?** I think we no longer need differentiations between released and unreleased versions, we can just find out by reading the Github release page.

azure-pipelines.yml
gunzip: What I meant is that we already have a build step here _into_ the deploy stage: https://github.com/pagopa/io-functions-admin/blob/master/azure-templates/deploy-steps.yml#L56 
So why do we want to make another build _before_ this one? (we build twice currently)
If the step inside deploy stage fails the deploy task stops anyway, isn't it?

azure-templates/make-release-steps.yml
gunzip: imho it's useful to have some pattern to detect release tags (ie. to trigger other events filtering on those tags)

azure-pipelines.yml
raicastino: I do not think it's possible. I know there is a way to share variables across multiple pipelines using variable group but the pipelines must be located in the same project (i.e. not at organization level). 

By the way, maybe it could be possible to use a custom solution to get this same result (e.g. just reading a well-known public file available on GitHub, e.g. https://github.com/pagopa/io-pagopa-commons/xxx)...

gunzip: if they're fixed we can consider hardcoding them into the .yml (and distribute the files using code generation)

azure-pipelines.yml
balanza: The deploy stage happens after the release is created, hence if it fails we have a _broken_ release to be manually deleted.

Test and build stages are there because we want some static checks before committing into a new release. I see that this leads to building twice, still I think it's the only way to cope with all the scenarios:
* broken build must not be released
* the code with the updated version must be deployed

azure-pipelines.yml
balanza: I think they're just fine as variables as we may need to re-configure them and that should not impact source code.

Moreover, connection is an information that really has meaning only inside Azure DevOps settings and email/username are loosely bound to that.

Given that pipelines have to be configured anyway, I'd not over-engineer on this and just use variables.

pasqualedevita: Maybe we can add a default hardcoded value here and pass an override parameter in azure-pipelines.yml only when the value is different.

https://github.com/pagopa/io-functions-admin/blob/175476527-release-pipeline/azure-templates/make-release-steps.yml#L12

gunzip:
why draft?

azure-pipelines.yml
pasqualedevita: It seems not possibile set a variable on organization level.

azure-pipelines.yml
gunzip: what if we move the release part after a successful build in the deploy stage?

balanza: The artefact we deploy would not have the version bumped. Apart from not referencing the correct commit in history, we would face the following:
* for services, we would not expose the correct version on `/info` endpoint
* for packages, we would ship the wrong version (or we would even be unable to deploy)

balanza:
> why draft?

Because we still need some refinement to make it robust

azure-pipelines.yml
pasqualedevita: I think that deploy step must be changed to deploy the released version on github.

balanza:
I updated the PR description with the updated changes and intended usage.

azure-templates/deploy-steps.yml
```diff
@@ -51,11 +51,17 @@ parameters:
...
+  # the branch, tag or commit to deploy
+  - name: 'git_reference'
+    type: string
+    default: $(Build.SourceVersion)
     
 steps:
   - template: ./make-build-steps.yml
     parameters:
       make: predeploy_build
+      git_reference: ${{ parameters.git_reference }}
```
pasqualedevita, line 64: We can improve deploy step downloading the release from github instead make the third build.
Maybe this will be done in next PR.

pasqualedevita (approved):

balanza:
Back on draft as I noticed errors composing the changelog

azure-templates/deploy-steps.yml
balanza: Next PR +1

balanza:
@gunzip I've added a section about version bump and changelog generation in the PR description, please have a look.

gunzip (approved):

azure-pipelines.yml
```diff
@@ -233,3 +287,10 @@ stages:
               azureSubscription: '$(PRODUCTION_AZURE_SUBSCRIPTION)'
               resourceGroupName: '$(PRODUCTION_RESOURCE_GROUP_NAME)'
               appName: '$(PRODUCTION_APP_NAME)'
+              # On the assumption that this stage is executed only when Relase stage is, 
+              #  with this parameter we set the reference the deploy script must pull changes from. 
+              # The branch/tag name is calculated from the source branch
+              #  ex: Build.SourceBranch=refs/heads/master --> master
+              #  ex: Build.SourceBranch=refs/tags/v1.2.3-RELEASE --> v1.2.3-RELEASE
+              git_reference: ${{ replace(replace(variables['Build.SourceBranch'], 'refs/tags/', ''), 'refs/heads/', '') }}
```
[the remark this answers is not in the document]
balanza: good point, I actually renamed it to be snake cased. It sounds more system-ish 😅 
I'll make a round of renaming to have all variables at the same notation.

balanza:
@gunzip @pasqualedevita @raicastino with latest commits I added a refinement that I'd like you to review. 

In brief: instead of evaluating conditions when the stage is executed, they are evaluated upfront when the pipeline is loaded. Results: **stages that won't run aren't shown**.
Basically, i put an if that includes `Release`, `Deploy_test` and `Deploy_production` stages

The benefit I see is we know earlier if a stage is going to be skipped (you can cancel pipeline asap if it's an error).

What do you think? If it's good practice, I can refine it a bit. If it's not, we can just revert latest commits and this PR is fine

gunzip:
it is not totally clear, how can the `succeeded()` condition be evalutated upfront?

balanza:
> it is not totally clear, how can the `succeeded()` condition be evalutated upfront?

doesn't `succeeded()` mean "so far so good"? If so, we're backed by the fact that if either `Build` or `Test` stage fail, the pipeline stops. Does it?

balanza:
> @gunzip @pasqualedevita @raicastino with latest commits I added a refinement that I'd like you to review.
> 
> In brief: instead of evaluating conditions when the stage is executed, they are evaluated upfront when the pipeline is loaded. Results: **stages that won't run aren't shown**.
> Basically, i put an if that includes `Release`, `Deploy_test` and `Deploy_production` stages
> 
> The benefit I see is we know earlier if a stage is going to be skipped (you can cancel pipeline asap if it's an error).
> 
> What do you think? If it's good practice, I can refine it a bit. If it's not, we can just revert latest commits and this PR is fine

Son in the end I had to revert this because Azure DevOps somehow doesn't handle multiple conditions as I would expected. I haven't found evidence of this on the internet, but I struggled a bit and I decided that it wasn't worth the effort.

In a nutshell, these work:
```yaml
${{ if eq(variables['Build.SourceBranch'], 'refs/heads/master') }}:
${{ if eq(variables['DO_RELEASE'], true) }}:
```
while these don't:
```yaml
# nested conditions
${{ if and( eq(variables['Build.SourceBranch'], 'refs/heads/master'), eq(variables['DO_RELEASE'], true) ) }}:

# nested ifs
${{ if eq(variables['Build.SourceBranch'], 'refs/heads/master') }}:
    ${{ if eq(variables['DO_RELEASE'], true) }}:
```

PR #110, opened 2020-11-25
[#175892280] Update max body size for logo update

infantesimone:
This PR update max body size to 50 mb for the functions:
- UploadOrganizationLogo
- UploadServiceLogo

### DIFF
--- a/UploadOrganizationLogo/index.ts
+++ b/UploadOrganizationLogo/index.ts
@@ -32,6 +32,7 @@ winston.add(contextTransport);
 const app = express();
 secureExpressApp(app);
 
+// Setup max body size
 app.use(bodyParser.json({ limit: "5mb" }));
 
 // Add express route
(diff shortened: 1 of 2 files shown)


UploadOrganizationLogo/index.ts
```diff
@@ -30,6 +32,8 @@ winston.add(contextTransport);
 const app = express();
 secureExpressApp(app);
 
+app.use(bodyParser.json({ limit: "50mb" }));
```
gunzip, line 35: is this a reasonable limit? (what's the expected max size? the current one?)

AleDore (commented):
LGTM @gunzip ?

UploadOrganizationLogo/index.ts
AleDore: Navigating through [io-services-metadata](https://github.com/pagopa/io-services-metadata/tree/master/logos) it seems that the expected max size is 1 Mb. With a conservative approach we can define a limit up to 5-10 MB.

gunzip: yes, can we update to 5mb?

infantesimone: ok 👍

AleDore (approved):
LGTM

balanza:
We are experiencing some troubles with PR's automatic check. We close this PR and we open a new one, let's see if this solves the issue.

PR #130, opened 2021-02-19
Add postbuild dependencies check

gquadrati:
In order to avoid missing dependencies after _yarn install --prod_ command, a postbuild check has been added to prompt all dependencies not listed in package.json


package.json
```diff
@@ -4,6 +4,7 @@
   "license": "MIT",
   "scripts": {
     "prebuild": "shx rm -rf dist && yarn generate",
+    "postbuild": "npx dependency-check package.json --missing ./dist/**/*.js",
```
balanza, line 7: Avoid the use of dev dependencies

```suggestion
    "postbuild": "npx dependency-check package.json --no-dev --missing ./dist/**/*.js",
```

balanza (changes requested):

package.json
gquadrati: using _--no-dev_ we see another dependency: _danger-plugin-digitalcitizenship_ 
This is used within Dangerfile.js
Is it correct to have this file in _dist/_ fodler?

tsconfig.json
```diff
@@ -12,6 +12,7 @@
     "__mocks__",
     "dist",
     "node_modules",
-    "**/__tests__/*"
+    "**/__tests__/*",
+    "Dangerfile.ts"
```
gquadrati, line 16: @balanza  do you think is correct to exclude this file from tsconfig.json?

balanza:
Ok, let's say we are fine with this check. We now have a failing check, and we must fix the build ;)

gquadrati:
> Ok, let's say we are fine with this check. We now have a failing check, and we must fix the build ;)

I wanted to try the Dangerfile before adding dependencies :) I'm doing it

balanza (approved):

PR #132, opened 2021-04-07
[#ENTE-20] Fix authorized CIDRs filter

BurnedMarshal:
#### List of Changes
- Array filter on the `retrievedServiceToApiService` method use the right CIDR decoder.

#### Motivation and Context
The CIDR type decoder from `italia-ts-commons` uses a different regex from the CIDR type decoder defined into the Service model and used on the OpenAPI specs.

- not tested

#### Types of changes

- [ ] Chore
- [X] Bug fix (non-breaking change which fixes an issue)


pasqualedevita (approved):
LGTM

balanza:
Should we align such regex somehow?

BurnedMarshal:
> Should we align such regex somehow?

@balanza Modify the used regex for models is dangerous, and have an impact on:
- OpenAPI specs that use that regex
- Models
Moreover changing regex on _model decoder_ make all the unaligned documents inaccessible through the code.
We can modify the `italia-ts-commons` CIDR regex, but I think that is a general definition that could be useful in other scenarios.

PR #145, opened 2021-05-21
[#IP-153] Recovery of pending failed user data processing

michaeldisaro:
#### List of Changes
- Added functions and orchestrator to find old failed user data processing requests

#### Motivation and Context
We want to find old failed user data processing requests and track their failure reason.

It has been tested by performing:
- yarn install --frozen-lockfile
- yarn build
- yarn test
- func host start

#### Types of changes

#### Checklist:


michaeldisaro:
@balanza Here are the basics for IP-153

UserDataProcessingFindFailureReasonActivity/handler.ts
```diff
@@ -0,0 +1,133 @@
...
+      // create failed orchestrator id based on choice
+      const failedUserDataProcessingOrchestratorId =
+        choice == UserDataProcessingChoiceEnum.DELETE
+          ? makeDeleteOrchestratorId(fiscalCode)
+          : makeDownloadOrchestratorId(fiscalCode);
+
+      return tryCatch(
+        () => {
+          // get the status of the failed orchestrator
+          return client.getStatus(failedUserDataProcessingOrchestratorId);
```
balanza, line 107: I don't understand why we need to read the actual status here. 
Besides of that, when using comment try to tell the _why_, the reason behind the choice you made; the _how_ is already told by code itself ;)

UserDataProcessingFindFailureReasonActivity/handler.ts
```diff
@@ -0,0 +1,133 @@
...
+        },
+        () => {
+          return ActivityResultFailure.encode({
+            kind: "NOT_FOUND_FAILURE"
+          });
+        }
+      );
+    })
+    .fold(identity, orchestratorStatus => {
+      return fromNullable("No reason found")(orchestratorStatus.output)
```
balanza, line 117: What kind of output do you expect?

UserDataProcessingSaveFailureReasonActivity/handler.ts
```diff
@@ -0,0 +1,152 @@
+/**
```
balanza, line 1: Is this activity actually used? Also, whoudn't `SetUserDataProcessingStatusActivity` be sufficient?

michaeldisaro: No, this is not used actually. In the end I decided to use SetUserDataProcessingStatusActivity and forgot to delete this.

UserDataProcessingFindFailureReasonActivity/handler.ts
michaeldisaro: I typed that output and decoded it, but in the end I thought that I needed a "string" reason that actually was saved in that "output" property. I take the string and save it as it is in the new reason. Do you prefer that I decode it as an ActivityFailure and then save a custom assembled string like "kind: reason"? Does this gives us any added value? What if for some strage reason there is not an ActivityFailure in that property and we cannot decode the string? Isn't better to have that string saved directly to the reason property?

UserDataProcessingFindFailureReasonActivity/handler.ts
michaeldisaro: Because I found that the orchestrator status contains its output. For old failed user data processing requests this output has the reason of the failure, or at least I think so based on data available to me (some tests done locally). What do you think about? Should I look elsewhere to get the reason?

UserDataProcessingProcessFailedRecords/handler.ts
```diff
@@ -0,0 +1,141 @@
...
+              await startUserDataProcessingRecoveryOrchestrator(context, udp);
+            });
+          });
+        }
+      })
+      .fold<IGetFailedUserDataProcessingHandlerResult>(
+        e => ResponseErrorQuery(`COSMOS|ERROR|${e.error}`, e),
+        s =>
+          ResponseSuccessAccepted(
+            "SUCCESS|Processing failed user data processing requests"
```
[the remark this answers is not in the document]
michaeldisaro: Ok, this was also a request of the issue. I'd have implemented after this initial flow. This is why this is actually a draft.

UserDataProcessingProcessFailedRecords/handler.ts
```diff
@@ -0,0 +1,141 @@
...
+): Promise<string> => {
+  const dfClient = df.getClient(context);
+  context.log.info(
+    `${logPrefix}: starting UserDataProcessingRecoveryOrchestrator for ${processable.choice}-${processable.fiscalCode}`
+  );
+  const orchestratorId = makeOrchestratorId(
+    processable.choice,
+    processable.fiscalCode
+  );
+  return startOrchestrator(
```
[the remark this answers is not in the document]
michaeldisaro: Ok, seems legit.

balanza (approved):

PR #146, opened 2021-06-04
[#IP-153] Fixed found reason with stringify to avoid decoding errors

michaeldisaro:


gquadrati (approved):
LGTM

UserDataProcessingFindFailureReasonActivity/__tests__/handler.test.ts
```diff
@@ -62,7 +62,7 @@ describe("UserDataProcessingFindFailureReasonActivity", () => {
     expect(decodedResult.isRight()).toBe(true);
     expect(decodedResult.value).toEqual({
       kind: "SUCCESS",
-      value: failedOrchestratorOutput
+      value: JSON.stringify(failedOrchestratorOutput, (key, value) => value)
```
gquadrati, line 65: Why the need of the replacer? Couldn't you just use `JSON.stringify(failedOrchestratorOutput)`?

michaeldisaro: I put that to avoid some [Object: object] result. That replacer will just take the values and convert them. Maybe I could have done a `${key}: ${value}`interpolation, but I think the values could be enough.

PR #156, opened 2021-07-23
[#IP-154] UserDataProcessingDeleteOrchestrator can manage previous failed requests if they are resumed

michaeldisaro:
#### List of Changes
- Added activity to check if a request has previously failed
- Added a spy on context mock
- Changed the logic of UserDataProcessingDeleteOrchestrator
- Added more tests

#### Motivation and Context
We want to run a recovery on FAILED delete requests by setting them to PENDING again and letting the orchestrator do the job. 
There are 3 points to respect:
- No grace period is waited for failed requests
- No email is sent for failed requests
- Everything must work as expected for new request

It has been tested by performing:
- yarn install --frozen-lockfile
- yarn build
- yarn lint
- yarn test
- tested on io-mock for the part regarding the new called activity to verify it's being called correctly

#### Types of changes

#### Checklist:

### DIFF
--- a/IsFailedUserDataProcessingActivity/handler.ts
+++ b/IsFailedUserDataProcessingActivity/handler.ts
@@ -1,0 +1,96 @@
+import { Context } from "@azure/functions";
+import { ServiceResponse, TableService } from "azure-storage";
+import { fromOption } from "fp-ts/lib/Either";
+import { NonEmptyString, FiscalCode } from "@pagopa/ts-commons/lib/strings";
+import { UserDataProcessingChoice } from "@pagopa/io-functions-commons/dist/generated/definitions/UserDataProcessingChoice";
+import { fromEither, tryCatch } from "fp-ts/lib/TaskEither";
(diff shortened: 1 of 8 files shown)


IsFailedUserDataProcessingActivity/handler.ts
```diff
@@ -0,0 +1,96 @@
...
+              failedUserDataProcessingTable,
+              i.choice,
+              i.fiscalCode,
+              null,
+              (error: Error, result: TableEntry, response: ServiceResponse) =>
+                response.isSuccessful
+                  ? resolve(some(result))
+                  : response.statusCode === 404
+                  ? resolve(none)
+                  : reject(error)
```
AleDore, line 69: Can we try to `taskify` this without manage explicitly a Promise?

michaeldisaro: Yes, I'd like that. Would you like to have a pair session?

AleDore (approved):
LGTM

PR #164, opened 2021-08-30
[#IP-374] Ensure password is shown in message

balanza:
#### Motivation and Context

We had evidence that some user, after claiming access on their data, received a message containing a password invalid for the associated data bundle.
We found out password was actually correct, but mis-rendered: when it contains two or more `*`, such characters would be considered as an italic formatting instruction in the markdown context, hence  they would not be shown to the user.

We remove `*` from the char collection. Also, we wrapped password value for escaping, to avoid similar issues to occur in the future

#### List of Changes

* Remove `*` from charset in https://github.com/pagopa/io-functions-admin/commit/1a30e0b7a942a9717d1f4de6da82d0725b3a1010
* Wrap password with back-ticks in https://github.com/pagopa/io-functions-admin/commit/ddde8af2436a68290e144fda20a8865d6e84be76

Unit tests added

#### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

#### Checklist:


BurnedMarshal (approved):
LGTM

PR #173, opened 2021-11-10
Add bulk move subscription owner

gquadrati:
#### List of Changes

Pair with @LoFrance 

Add new `POST - adm/subscriptions/ownerships` endpoint

#### Motivation and Context

We need a programmatically way of moving a list of subscriptions from an user to anther, to help backoffice operations.

Unit tests

#### Types of changes

#### Checklist:

### DIFF
--- a/openapi/index.yaml
+++ b/openapi/index.yaml
@@ -469,7 +469,7 @@ paths:
         "500":
           description: Internal server error
   /actions/subscriptions/ownerships:
-    patch:
+    post:
       summary: Update Subscriptions owner
       description: Update Ownership for a group of subscriptions
       operationId: updateSubscriptionOwner
--- a/openapi/index.yaml
+++ b/openapi/index.yaml
@@ -468,6 +468,37 @@ paths:
           description: Resource (User or Product) not found
         "500":
           description: Internal server error
+  /subscriptions/{subscription_id}/owner:
+    patch:
+      summary: Update Subscription owner
+      description: Update Subscription owner
+      operationId: updateSubscriptionOwner
+      parameters:
+        - name: subscription_id
+          in: path
+          type: string
+          required: true
+          description: The id of the Subscription
+        - name: body
+          in: body
+          required: true
+          schema:
+            $ref: "#/definitions/UpdateSubscriptionOwnerPayload"
+      responses:
+        "200":
+          description: Subscription updated
+        "400":
+          description: Bad request
+        "401":
+          description: Unauthorized
+        "403":
+          description: Forbidden
+        "404":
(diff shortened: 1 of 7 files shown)


openapi/index.yaml
```diff
@@ -468,6 +468,32 @@ paths:
           description: Resource (User or Product) not found
         "500":
           description: Internal server error
+  /subscriptions/move-owner:
```
gunzip, line 471: it's not clear to me if with 'subscription owner' we are here referring to the 'user' resource
which has its own endpoints and is used elsewhere 
(why not just use user instead of 'subscription owner' in case ?)

since we are dealing with multiple subscriptions, I guess a better description would be 
'Move a list of subscriptions from an owner to another' (the one you've used above in the code :-)
and use the plural form in the operation name (ie. MoveUserSubscriptions if they all belong
to the same user).

-----

nit: it's better to avoid rpc like endpoint (verbs) and try to use resource names instead, you can
use  something like `/actions/subscriptions` and pass the operation name into the payload
(if one day you'll need a `/subscription/{subscription_id}` that would conflict with
the 'move-owner' endpoint when using the proposed layout)

nit: another "optimization" (but this is not needed, just consider this in case you want to expand the set
of bulk operations) is to pass the operation name for each item in the payload, so you can mix multiple
operations in the same payload (same bulk action, different operations).

gquadrati: Thanks, we were unhappy about the path, I appreciate you're feedbacks.
I like the idea of having an /action route, but I prefer to pass the operation name as path parameter so we can have one   Azure Function per each operation.

What do you think of `/action/move-subscriptions`?

gunzip: `/actions/subscriptions/ownerships` ?

gunzip: or `/actions/subscriptions/move` ?

gquadrati: Perfect, thanks! 
And is it ok to have it as `patch` or we should use `put`?

gunzip: I'm not biased, I think POST (since it's conceptually a new move operation) would be ok

gunzip: to clarify, I'd gone with 'PATH' just in case the URL were '/subscriptions' and the op specified into payload

gquadrati: ok, thank you! :)

gquadrati: Added in [f4ce14d](https://github.com/pagopa/io-functions-admin/pull/173/commits/f4ce14d018f8d465df619537e4d7f4149dafc7f0)

PR #202, opened 2022-12-09
[#IOPSC-118] Respond with 'too many requests' status code when creating too much subscriptions for the same account

balanza:
#### List of Changes

See commit list

#### Motivation and Context

Azure API Management may fail to create too many subscriptions for the same account. According to Microsoft support, this is due to its internal optimistic consistency mechanism. Such failures do not depend on us or on a specific traffic pattern and are non-deterministic from our perspective; nevertheless, they are treated as generic server errors 500 and are hard to be monitored apart and react to differently.

To improve our users' experience we decided to:
* retry up to 3 times on the `subscriptions.createOrUpdate` operation when a 412 error is returned, so our users do not have to care about anything at the cost of a slightly-higher response time
* respond with 429 if retries do not solve, so that they can regulate request rate themselves (it will also be helpful for us as we can better monitor failures)

#### Types of changes

- [ ] Chore (nothing changes from a user perspective)

#### Checklist:

### DIFF
--- a/CreateSubscription/__tests__/handler.test.ts
+++ b/CreateSubscription/__tests__/handler.test.ts
@@ -6,6 +6,7 @@ import {
   SubscriptionContract,
   UserContract
 } from "@azure/arm-apimanagement/esm/models";
+import { RestError } from "@azure/ms-rest-js";
 import * as E from "fp-ts/lib/Either";
 import * as TE from "fp-ts/lib/TaskEither";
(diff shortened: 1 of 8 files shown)


balanza:
cc @rodoviti @giuseppedipinto @stefanodoc

LoFrance (changes requested):

PR #211, opened 2023-01-30
[#IOPAE-141]  Add GetSubscription admin API

giuseppedipinto:
Added new **GetSubscription** admin API

#### List of Changes

- Added new GetSubscription API

#### Motivation and Context

Used to retrieve a single subscription starting from a subscriptionId

#### Types of changes

#### Checklist:

### DIFF
--- a/GetSubscription/handler.ts
+++ b/GetSubscription/handler.ts
@@ -1,0 +1,113 @@
+import { Context } from "@azure/functions";
+import * as express from "express";
+import * as TE from "fp-ts/lib/TaskEither";
+import {
+  AzureApiAuthMiddleware,
+  IAzureApiAuthorization,
+  UserGroup
+} from "@pagopa/io-functions-commons/dist/src/utils/middlewares/azure_api_auth";
+import { ContextMiddleware } from "@pagopa/io-functions-commons/dist/src/utils/middlewares/context_middleware";
+import { RequiredParamMiddleware } from "@pagopa/io-functions-commons/dist/src/utils/middlewares/required_param";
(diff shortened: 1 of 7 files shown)


GetSubscription/handler.ts
```diff
@@ -0,0 +1,113 @@
...
+            subscriptionId
+          ),
+          TE.mapLeft(mapApimRestError("Subscription")),
+          TE.map(subscription =>
+            ResponseSuccessJson({
+              id: subscription.id,
+              owner_id: parseOwnerIdFullPath(
+                subscription.ownerId as NonEmptyString
+              ),
+              primary_key: subscription.primaryKey,
```
balanza, line 78: Do we need to expose `primary_key` and `secondary_key` in this API? 
If not striclty required, I'd rather not expose such information. 
If it's required it's still ok, let just spend some time reasoning about how we ensure nobody is getting an apykey of a subscription it doesn't own.

(I see this API is protected at admin level, but for the _least data principle_ we might be mimimize the data we expose.)

balanza (changes requested):
just a clarification

GetSubscription/handler.ts
giuseppedipinto: In our specific case, basically, only the ownerId of the subscription is needed.
But since it is a generic API we decided not to alter the Subscription interface except for the part that not concerning the concept of subscription itself _(i.e. the service)_.
We can decide together whether to restrict it to only the data necessary for our purpose or whether to expand the amount of data to be displayed in continuity with the current **Subscription** interface.

balanza (approved):

PR #225, opened 2023-07-04
[#IOPAE-431] Delete cmsTag field before updating a service

giamma1295:
The new `io-services-cms` has a functionality to sync back and forth the services between the new and the legacy containers:

- When a service is created/updated using the new APIs it is also written into the legacy container and marked with a field `cmsTag`
- When a service is created/updated using the old APIs a CosmosDBTrigger Azure Function will intercept it an write it into the new container but 
     only if the `cmsTag` field is not present, so when a service is updated using the old APIs the "cmsTag" field needs to be removed.
#### List of Changes

- Remove `cmsTag` before updating a service

Unit Test
#### Types of changes

#### Checklist:


giuseppedipinto (approved):
LGTM

PR #227, opened 2024-01-05
[IOPID-1032] Add SanitizeProfileEmail to make profiles compliant with the unique email enforcement

lucacavallaro:
Depends on https://github.com/pagopa/io-infra/pull/814

#### List of Changes

Add `SanitizeProfileEmail` function that, given a profile to sanitize identified by `fiscalCode` and `email` creates a new profile version with `isEmailValidated = false`

The following sequence diagram illustrates the procedure

image

##### What if?

> What if there are errors on retrieving a profile or update?

The function fails with an error and does not consume the "profile item" from the queue, after **five** attempts the profile to sanitize will be moved to `profiles-to-sanitize-poison` queue

> What if a profile (identified by `fiscalCode`) is processed two times?

The produce will create a new profile version with `isEmailValidated = false` only if the latest version of the profile has `isEmailValidated = true`, so the second time the update will be skipped.

#### Motivation and Context

This procedure is required to sanitize user profiles with duplicate email addresses after the release of the unique email enforcement feature

`io-mock` + unit test

#### Types of changes

#### Checklist:


package.json
```diff
@@ -36,19 +36,20 @@
     "@types/jest": "^24.9.1",
     "@types/upng-js": "^2.1.1",
     "auto-changelog": "^2.3.0",
+    "azure-functions-core-tools": "^4.0.5455",
```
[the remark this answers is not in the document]
lucacavallaro: yes, I have the same effect with `yarn 1`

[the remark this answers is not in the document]
lucacavallaro: but it should explicitly listed, since there are script that uses this tool

SanitizeProfileEmail/main.ts
```diff
@@ -0,0 +1,27 @@
...
+const profilesContainer = cosmosdbInstance.container(PROFILE_COLLECTION_NAME);
+
+const profileModel = new ProfileModel(profilesContainer);
+
+const createSanitizeProfileEmailsFunction = azureFunction(
+  H.of(sanitizeProfileEmail)
+);
+
+export default createSanitizeProfileEmailsFunction({
+  inputDecoder: ProfileToSanitize,
```
gquadrati, line 25: What should happen in case the input does not satisfy the decoder?

SanitizeProfileEmail/index.ts
```diff
@@ -0,0 +1,82 @@
...
+  );
+
+const setProfileEmailAsNotValidated = (profile: RetrievedProfile) => (
+  r: IProfileModel
+): TE.TaskEither<Error, void> =>
+  pipe(
+    r.profileModel.update({
+      ...profile,
+      isEmailValidated: false
+    }),
```
gquadrati, line 59: Also update `isEmailEnabled` based on `profile._ts`, as defined in DR.

gquadrati (changes requested):
Some small comments

env.example
```diff
@@ -62,4 +62,8 @@ FailedUserDataProcessingStorageConnection=<STORAGE_CONNECTION_STRING>
 FAILED_USER_DATA_PROCESSING_TABLE=FailedUserDataProcessing
 
 LOCKED_PROFILES_STORAGE_CONNECTION_STRING=<STORAGE_CONNECTION_STRING>
-LOCKED_PROFILES_TABLE_NAME=lockedprofiles
\ No newline at end of file
+LOCKED_PROFILES_TABLE_NAME=lockedprofiles
+
+### UNIQUE EMAIL ENFORCEMENT
+SanitizeUserProfileQueueName = profiles-to-sanitize
+CitizenAuthStorageConnection = <STORAGE-CONNECTION-STRING>
```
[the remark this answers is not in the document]
lucacavallaro: sure!

SanitizeProfileEmail/main.ts
lucacavallaro: the function exits with a `ValidationError`, after five retries the malformed input will be moved to `profile-to-sanitize-poison` queue

(the same appens with unhandled errors in the controller)

SanitizeProfileEmail/__tests__/handler.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    await Promise.all(
+      profiles.map(p =>
+        sanitizeProfileEmail(p)({ profileModel, logger: ConsoleLogger })()
+      )
+    );
+
+    expect(MockedProfileModel.prototype.update).toBeCalledTimes(1);
+    expect(MockedProfileModel.prototype.update).toBeCalledWith(
+      expect.objectContaining({ isEmailValidated: false })
```
gquadrati, line 112: Since we are running all tests at once, check the fiscalCode too (you need to update the `findLastVersionByModelId` mock otherwise it will fail).

```suggestion
      expect.objectContaining({
        isEmailValidated: false,
        fiscalCode: mocks.fiscalCodes.TO_SANITIZE
      })
    );
```

SanitizeProfileEmail/__tests__/handler.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+      },
+      {
+        email: mocks.email,
+        fiscalCode: mocks.fiscalCodes.EMAIL_NOT_VALIDATED
+      },
+      {
+        email: mocks.email,
+        fiscalCode: mocks.fiscalCodes.NOT_FOUND
+      }
+    ];
```
gquadrati, line 102: Add a test for checking `isEmailEnabled`update

gquadrati (changes requested):

gquadrati (approved):

PR #229, opened 2024-01-18
[#IOPID-1030] Added delete profileEmail logic to DeleteUserDataActivity

silvicir:
#### List of Changes
Updated `DeleteUserDataActivity` function with the cancellation of `profileEmails` entity as last operation

#### Motivation and Context
Now we have the `profileEmails` table storage populated with the validated email for each fiscal code. When a user requires their data deletion, the respective entity must be cancelled from the `profileEmails`. Then the email can be selected and validated again by someone else.

Through unit tests and `io-mock`

#### Types of changes

#### Checklist:


DeleteUserDataActivity/backupAndDelete.ts
```diff
@@ -155,13 +203,17 @@ const backupAndDeleteProfile = ({
...
+  profileEmailsRepository,
+  lastValidatedEmail
 }: {
   readonly authenticationLockService: AuthenticationLockService;
   readonly profileModel: ProfileDeletableModel;
   readonly userDataBackup: IBlobServiceInfo;
   readonly servicePreferencesModel: ServicePreferencesDeletableModel;
   readonly fiscalCode: FiscalCode;
+  readonly profileEmailsRepository: IProfileEmailWriter;
+  readonly lastValidatedEmail?: EmailString;
```
gquadrati, line 216: Can we use `O.Option<EmailString>` instead?

DeleteUserDataActivity/backupAndDelete.ts
```diff
@@ -622,14 +687,25 @@ export const backupAndDeleteAllUserData = ({
...
+    TE.chainW(() =>
+      pipe(
+        getLastEmailIfValidated({ fiscalCode, profileModel }),
+        TE.fold(
+          () => TE.right(void 0),
+          lastValidatedEmail =>
+            TE.right(pipe(lastValidatedEmail, O.toUndefined))
+        ),
+        TE.chain(lastValidatedEmail =>
+          backupAndDeleteProfile({
```
gquadrati, line 699: A style suggestion: could we relocate `getLastEmailIfValidated` into `backupAndDeleteProfile` since it pertains to the profile?

DeleteUserDataActivity/backupAndDelete.ts
silvicir: it's better not to use `Option` in parameters. It's easier to access the argument and so it can be easily validated

gquadrati: Since `getLastEmailIfValidated` is returning an Option, we can avoid useless steps [here](https://github.com/pagopa/io-functions-admin/blob/84c3516a61eca6832ac3c4a721ad34b42e42225f/DeleteUserDataActivity/backupAndDelete.ts#L693)

silvicir: I put `getLastEmailIfValidated` into `backupAndDeleteProfile` as you preferred so I removed `lastValidatedEmail` from the parameters

gquadrati (approved):
LGTM, good job!

PR #231, opened 2024-03-05
[#IOPID-1450] Enable `strict` mode on repo

gquadrati:
#### Motivation and Context

Enable `strict` mode on this repo by incrementally enabling the following flags (see [here](https://www.typescriptlang.org/tsconfig#strict)):

- [x] noImplicitAny
- [x] strictNullChecks
- [x] noImplicitThis
- [x] strictFunctionTypes
- [x] strictPropertyInitialization
- [x] strictBindCallApply
- [x] alwaysStrict
- [x] useUnknownInCatchVariables

Since Strict mode is actually a combination of this six other flags.

Unit tests

#### Types of changes

#### Checklist:


DeleteUserDataActivity/utils.ts
```diff
@@ -11,6 +14,21 @@ import {
   QueryFailure
 } from "./types";
 
+// Cosmos Errors
+export enum CosmosErrorsTypes {
+  "OSMOS_EMPTY_RESPONSE" = "OSMOS_EMPTY_RESPONSE",
+  "OSMOS_CONFLICT_RESPONSE" = "OSMOS_CONFLICT_RESPONSE",
+  "OSMOS_DECODING_ERROR" = "OSMOS_DECODING_ERROR",
+  "OSMOS_ERROR_RESPONSE" = "OSMOS_ERROR_RESPONSE"
```
rodoviti, line 22: typo?!
```suggestion
  "COSMOS_EMPTY_RESPONSE" = "COSMOS_EMPTY_RESPONSE",
  "COSMOS_CONFLICT_RESPONSE" = "COSMOS_CONFLICT_RESPONSE",
  "COSMOS_DECODING_ERROR" = "COSMOS_DECODING_ERROR",
  "COSMOS_ERROR_RESPONSE" = "COSMOS_ERROR_RESPONSE"
```

gquadrati: Definitely 😓  Thanks!!

rodoviti (approved):
LGTM (for "services pov")

DeleteUserDataActivity/utils.ts
gquadrati: I've incorporated a test to verify the accuracy of enum elements by defining a tuple object containing all possible literal values of the `kind` property within `CosmosError`. By doing so, the test will signal failure upon any addition of a new 'kind' to `CosmosError`, prompting us to update the enum accordingly.
See [[#IOPID-1450] add tests for isCosmosErrors](https://github.com/pagopa/io-functions-admin/pull/231/commits/2e0f07e641fb16e12ad322da065762119eea0fb1)

DeleteUserDataActivity/__tests__/backupAndDelete.test.ts
```diff
@@ -37,12 +37,18 @@ const asyncIteratorOf = <T>(items: T[]): AsyncIterator<T[]> => {
       const value = data.shift();
       return {
         done: typeof value === "undefined",
-        value: [value]
+        value: [value!]
```
BurnedMarshal, line 40: Could be this unsafe? In tests I don't see particular problems.

DeleteUserDataActivity/backupAndDelete.ts
```diff
@@ -2,12 +2,11 @@ import * as crypto from "crypto";
 
 import { BlobService } from "azure-storage";
 import { sequenceT } from "fp-ts/lib/Apply";
-import * as A from "fp-ts/lib/Array";
+import * as RA from "fp-ts/lib/ReadonlyArray";
```
BurnedMarshal, line 5: In other files we use `ROA`. Maybe is better use the same name across all the files of the project.

GetFailedUserDataProcessing/handler.ts
```diff
@@ -62,6 +62,9 @@ export const GetFailedUserDataProcessingHandler = (
             failedUserDataProcessingTable,
             choice,
             fiscalCode,
+            // TODO: Refactor for using the new `@azure/data-tables` library
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore
             null,
```
BurnedMarshal, line 68: This code works because the type definition of the library is wrong or incomplete?

GetServices/handler.ts
```diff
@@ -83,7 +83,8 @@ export function GetServicesHandler(
                 // keep only the latest version
                 const isNewer =
                   !prev.has(curr.serviceId) ||
-                  curr.version > prev.get(curr.serviceId);
+                  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                  curr.version > prev.get(curr.serviceId)!;
```
BurnedMarshal, line 87: This is unsafe. The undefined or null value is ignored as if the `strict` option is still disabled.

BurnedMarshal (changes requested):
Part 1

GetServices/handler.ts
gquadrati: By construction, `prev.get(curr.serviceId)` will never be undefined for the following reasons:
- we've confirmed its existence in the initial part of the predicate, ensuring the `get` function returns a value
- `prev` is a map of NonNegativeInteger type values, which cannot be null

If this explanation suffices, I'll proceed to add a comment.

DeleteUserDataActivity/__tests__/backupAndDelete.test.ts
gquadrati: At the end, I preferred to rewrite the function with a simpler implementation :) See [[#IOPID-1450] refactor based on comments, part 1](https://github.com/pagopa/io-functions-admin/pull/231/commits/13289a70c86f945983ec6e17fee2ccff25ad2f77)

DeleteUserDataActivity/backupAndDelete.ts
gquadrati: Right, done in [[#IOPID-1450] refactor based on comments, part 1](https://github.com/pagopa/io-functions-admin/pull/231/commits/13289a70c86f945983ec6e17fee2ccff25ad2f77)

GetServices/handler.ts
BurnedMarshal: Depends on the previous row, if for a mistake `!prev.has(curr.serviceId)` become `prev.has(curr.serviceId)` the type checker doesn't show any error and the `undefined` result is possible. One of the main goals that TypeScript wants to achieve is to prevent errors.

What if we use something like this?
```typescript
pipe(
  prev.has(curr.serviceId),
  O.fromPredicate(identity),
  O.chainNullableK(() => prev.get(curr.serviceId)),
  O.fold(
    () => true,
    prevVersion => curr.version > prevVersion
  ),
  isNewer =>
    isNewer ? prev.set(curr.serviceId, curr.version) : prev
)
```

gquadrati: > Depends on the previous row, if for a mistake !prev.has(curr.serviceId) become prev.has(curr.serviceId) the type checker doesn't show any error and the undefined result is possible.

In this scenario, a test would fail, and I didn't want to change the code.

Anyway, I've updated it, thanks! [[#IOPID-1450] refactor based on comments, part 2](https://github.com/pagopa/io-functions-admin/pull/231/commits/d70ccaa3c845511168ae625d3497e20d15be64ad)

arcogabbo (approved):

PR #248, opened 2024-11-25
[#IOPID-2503] add masking preprocessor for appinsights

arcogabbo:
#### Motivation and Context

added preprocessor capability to applicationinsights

local run

#### Types of changes

#### Checklist:

### DIFF
--- a/utils/appinsights.ts
+++ b/utils/appinsights.ts
@@ -1,5 +1,6 @@
 import * as ai from "applicationinsights";
 import {
+  Data,
   EventTelemetry,
   ExceptionTelemetry
 } from "applicationinsights/out/Declarations/Contracts";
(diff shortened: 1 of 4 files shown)


utils/appinsights.ts
```diff
@@ -14,25 +15,69 @@ import { pipe } from "fp-ts/lib/function";
...
+    readonly [name: string]: unknown;
+  }
+): boolean => {
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const data = (envelope.data as Data<any>).baseData;
+    const userDataProcessingId = data.properties[USER_DATA_PROCESSING_ID_KEY];
+    if (
+      userDataProcessingId !== undefined &&
+      typeof userDataProcessingId === "string"
```
BurnedMarshal, line 31: What if we use `FiscalCode.is` instead?

BurnedMarshal (changes requested):

utils/appinsights.ts
BurnedMarshal: Sorry, the ID include the operation type, is not a plain CF

BurnedMarshal (approved):

PR #297, opened 2026-08-03
IOPLT-1919 -  Upgrade infra apim modules to azurerm v4

ansmatbitrock:
Upgrade the infra apim modules  to azurerm v4, to enable the porting of all the function admin resources from [`io-infra/common`](https://github.com/pagopa/io-infra/tree/main/src/common/_modules/function_admin) to the monorepo


infra/resources/_modules/apim/products.tf
```diff
@@ -1,5 +1,5 @@
 module "apim_v2_product_admin" {
-  source = "github.com/pagopa/terraform-azurerm-v3//api_management_product?ref=v8.27.0"
+  source = "github.com/pagopa/terraform-azurerm-v4//api_management_product?ref=v1.0.0"
```
BurnedMarshal, line 2: Is this version locked to `v1.0.0` instead of `v6.14.0` for some reason?

BurnedMarshal (approved):

infra/resources/_modules/apim/products.tf
ansmatbitrock: yes, there is only one tag for the module available:

https://github.com/pagopa/terraform-azurerm-v4/commit/e4bda3675d83fa04bb83373d88b9a3c1f358397d

